PR #13, opened 2016-07-26
Implement ITP, remove es6 classes

emschwartz:
This PR looks fairly ugly because I restarted work on this library intending for it to be a separate ITP module. However, we decided that `ilp-client` should just use ITP by default for now and we can change it to add other transport protocols when those are developed. This commit bundles all of the changes that were intended for that ITP-specific library.


docs/README.template.md
```diff
@@ -52,65 +47,110 @@ The [Universal Transport Protocol (UTP)](https://github.com/interledger/rfcs/blo
...
 ```
 
 ### Paying
 ```js
-import { Client } from 'ilp'
-const client = new Client({
+'use strict'
+
+const ILP = require('ilp')
+const sender = new sender({
```
justmoon: wat?

justmoon:
LGTM

PR #16, opened 2016-07-28
Expect full ILP packet & add disallowOverPayment

emschwartz:
The receiver should expect the full ilp packet (request) in the incoming transfer data field.

This also adds an option to require that incoming transfer amounts exactly match the requested amount.


src/lib/receiver.js
```diff
@@ -24,6 +24,7 @@ const BigNumber = require('bignumber.js')
  * @param  {ilp-core.Client} [opts.client] [ilp-core](https://github.com/interledger/js-ilp-core) Client, which can optionally be supplied instead of the previous options
  * @param  {Buffer} [opts.hmacKey=crypto.randomBytes(32)] 32-byte secret used for generating request conditions
  * @param  {Number} [opts.defaultRequestTimeout=30] Default time in seconds that requests will be valid for
+ * @param  {Boolean} [opts.disallowOverPayment=false] Require that incoming transfer amounts exactly match the requested amount
```
justmoon: I would reverse this: `allowOverPayment=false`.

I get that overpayment is theoretically good for the recipient, but in practice isn't it just going to mean that some error happened and if you don't decline the payment you're going to have to manually go in an fix it later?

At least in ITP it seems like overpayment would be an indicator of a serious problem and should be rejected. And if somebody truly needs it they can set `allowOverPayment`.

Also, I think it should be `allowOverpayment`, see [overpayment](http://dictionary.cambridge.org/dictionary/english/overpayment).

emschwartz: new commit changes this behavior

emschwartz:
@justmoon note that one result of requiring the incoming amount to exactly match the requested amount is that we'll need to round the amount given by the user to the receiving ledger's precision when creating the request. Otherwise, you could have a request that would be impossible to fulfill.

emschwartz:
Alternatively, we could allow transfer amounts that are greater than the requested amount because of rounding

justmoon:
> note that one result of requiring the incoming amount to exactly match the requested amount is that we'll need to round the amount given by the user to the receiving ledger's precision when creating the request. Otherwise, you could have a request that would be impossible to fulfill.

Good observation. Should be doable since the receiver is guaranteed to have a relationship to their ledger. We probably want to always round up to avoid weird exploits. (I can create money by depositing 0.00499 into a wallet a million times.)

LGTM

PR #20, opened 2016-08-09
Update to ilp-core@6.x, including ILP addresses and new plugin options

emschwartz:
Update to latest `ilp-core`
Use new request format:
```json
{
  "packet": { ... ILP Packet ... },
  "condition": "cc..."
}
```


sentientwaffle:
LGTM

PR #27, opened 2016-08-24
Add sender.quoteSourceAmount and receiver.getAddress

emschwartz:
This is needed to use js-ilp in SPSP and other higher level protocols where the sender can ask the
receiver for a payment request. Getting a quote for a fixed source amount is useful so the sender
can figure out the amount they should tell the receiver for the (fixed destination amount) request. This also enables the receiver to get their ILP address so that it can be communicated to the sender before the sender requests a fixed source amount quote.


justmoon:
Super-clean! LGTM

PR #31, opened 2016-08-26
fix: autoFulfillConditions: use paymentRequest for condition

vhpoet:
This PR does 2 things.

1. Fixes the condition generation in `autoFulfillConditions`.
2. Makes `paymentRequest` include `data` only if it exists.

### DIFF
--- a/src/lib/receiver.js
+++ b/src/lib/receiver.js
@@ -77,8 +77,11 @@ function createReceiver (opts) {
     const paymentRequest = {
       address: account + '.' + (params.id || uuid.v4()),
       amount: String(params.amount),
-      expires_at: params.expiresAt || moment().add(defaultRequestTimeout, 'seconds').toISOString(),
-      data: params.data
+      expires_at: params.expiresAt || moment().add(defaultRequestTimeout, 'seconds').toISOString()
+    }
+
+    if (params.data) {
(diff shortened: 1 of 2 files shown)


src/lib/receiver.js
```diff
@@ -118,10 +121,12 @@ function createReceiver (opts) {
     const paymentRequest = {
       address: packet.account,
       amount: packet.amount,
-      expires_at: packet.data.expires_at,
-      data: packet.data.data
+      expires_at: packet.data.expires_at
```
emschwartz: It would be even better if this were `packet.data && packet.data.expires_at`

src/lib/receiver.js
```diff
@@ -118,10 +121,12 @@ function createReceiver (opts) {
...
     }
 
+    if (packet.data.data) {
```
emschwartz: Should be `if (packet.data && packet.data.data) {`

emschwartz:
Should add a test or two to make sure this problem doesn't come back later

src/lib/receiver.js
vhpoet: in that case you would end up with a `{data: false}` which is still unnecessary.

vhpoet:
@emschwartz I think there are already tests for this, I've updated them.

PR #32, opened 2016-08-26
fix: use paymentRequest for condition

sentientwaffle:
https://github.com/interledger/js-ilp/pull/31 with fixes

### DIFF
--- a/test/receiverSpec.js
+++ b/test/receiverSpec.js
@@ -187,6 +187,21 @@ describe('Receiver Module', function () {
         expect(request.expires_at).to.equal('1970-01-01T00:00:30.000Z')
       })
 
+      it('should set the data if supplied', function () {
+        const request = this.receiver.createRequest({
+          amount: 10,
+          data: {foo: 'bar'}
+        })
+        expect(request.data).to.deep.equal({foo: 'bar'})
+      })
+
+      it('should not set the data if not supplied', function () {
+        const request = this.receiver.createRequest({
+          amount: 10
+        })
+        expect(request).to.not.have.keys('data')
+      })
+
       it.skip('should generate the condition from the request details', function () {
 
       })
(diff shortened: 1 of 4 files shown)


emschwartz:
It would probably be good to add a test to make sure this kind of bug doesn't come bak in the future

test/receiverSpec.js
```diff
@@ -288,6 +303,28 @@ describe('Receiver Module', function () {
           expect(results).to.deep.equal(['sent'])
         })
 
+        it('should ignore transfers without packet.data', function * () {
```
emschwartz: Actually I think it should be valid for a request to not include an expiry, but in this case the meat of the test is correct because it's really showing a case where the packet has been tampered with.

test/receiverSpec.js
```diff
@@ -288,6 +303,28 @@ describe('Receiver Module', function () {
...
+          delete this.transfer.data.ilp_header.data
+          const request = this.receiver.createRequest({
+            amount: 1,
+            id: '22e315dc-3f99-4f89-9914-1987ceaa906d'
+          })
+          const results = yield this.client.emitAsync('incoming_prepare', this.transfer)
+          expect(results).to.deep.equal(['condition-mismatch'])
+        })
+
+        it('should fulfill conditions of transfers packet.data.expires_at but no packet.data.data', function * () {
```
emschwartz: Hmm, it should fulfill requests that didn't have data in them originally but if the data is deleted from a request that should fail. Seems like this wording and the test should be changed a bit.

test/receiverSpec.js
sentientwaffle: Ah, right. This is really testing a mismatched request/transfer

test/receiverSpec.js
```diff
@@ -288,6 +303,28 @@ describe('Receiver Module', function () {
...
+          const request = this.receiver.createRequest({
+            amount: 1,
+            id: '22e315dc-3f99-4f89-9914-1987ceaa906d',
+            expiresAt: this.transfer.data.ilp_header.data.expires_at
+          })
+          delete this.transfer.data.ilp_header.data.data
```
emschwartz: If you want to test what happens when the packet is modified and the data is tampered with, you should include the data in the `createRequest` params and then delete it

sentientwaffle: No, this is just testing what happens when both the request and transfer don't have `ilp_header.data.data`.

emschwartz: Then maybe the test should just be called `should fulfill transfers corresponding to requests with no data included`

emschwartz:
LGTM

PR #33, opened 2016-08-29
Add sender.quoteDestinationAmount

emschwartz:
sender.quoteRequest could theoretically be used instead of quoting the destination amount, but this
unnecessarily burdens the receiver in cases where the sender may ask for many quotes before deciding
to send the payment. this adds a function to request a purely informational quote for a destination
amount

requested in https://github.com/interledger/five-bells-wallet/pull/25


justmoon:
LGTM

PR #34, opened 2016-08-29
Payment request-specific events for fulfilled requests

emschwartz:
allow users to add event listeners for specific payment requests such as receiver.on('incoming:requestid'). also allow wildcards such that receiver.on('incoming:appid.*') will match the 'incoming:appid.requestid' event


justmoon:
LGTM

PR #38, opened 2016-08-31
Add receiver.stopListening, remove listeners from sender too

emschwartz:
this is to help users of this library clean up when they no longer need the ILP client.
receiver.stopListening is an explicit command to remove the listeners and disconnect the underlying
client. for the sender, listeners should be removed once an outgoing payment is fulfilled or times
out.

resolves https://github.com/interledger/js-ilp/issues/36


sentientwaffle:
LGTM

PR #42, opened 2016-09-11
Handle request amounts w/ too many decimal places and significant digits

emschwartz:
By default it will throw an error if you try to create a payment request with an amount that has more decimal places or significant digits than the ledger supports.

Also adds the receiver option `roundRequestAmounts`, which can be set to `UP`, `DOWN`, `HALF_DOWN`, etc to round requests amounts with too many decimal places.

Resolves https://github.com/interledger/js-ilp/issues/17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /node_modules/
 /coverage/
 docs/*.intermediate.md
+*.log
+
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ of transfers paying for the payment requests created by the Receiver.
 | [opts.hmacKey] | <code>Buffer</code> | <code>crypto.randomBytes(32)</code> | 32-byte secret used for generating request conditions |
 | [opts.defaultRequestTimeout] | <code>Number</code> | <code>30</code> | Default time in seconds that requests will be valid for |
 | [opts.allowOverPayment] | <code>Boolean</code> | <code>false</code> | Allow transfers where the amount is greater than requested |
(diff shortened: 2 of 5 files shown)


emschwartz:
Now I'm wondering if the rounding mode should be an option on the receiver, a parameter for `createRequest`, or both.

sentientwaffle:
I think a parameter of `createRequest` is better -- it allows for more granularity in selecting the rounding mode.

emschwartz:
The argument against that would be that you're passing in the amount right there so you should just round yourself if you want it to be rounded. To support that behavior we could add a `getInfo` or `getScale` method so you'd know what to round it to

sentientwaffle:
Ah, I didn't think of adding those methods. I think a parameter of `createRequest` is still simpler, though -- it just adds a parameter instead of both a parameter and a function.

emschwartz:
You wouldn't need a parameter and a function. We can either make the user do the rounding themselves, or we can do it for them with a per-receiver or a per-request parameter (or both, such that the former would be the default).

The idea with adding `getInfo` would be to make the user do the rounding themselves so whatever you give to `createRequest` is what it would be. An advantage of this is then you know exactly what you're rounding to. I could potentially see some unhappy customers realizing that the round `'UP'` method was rounding up more than they wanted because the number of decimal places allowed was fewer than they expected (e.g. 0.0001 getting rounded to 0.01).

emschwartz:
Talked with @justmoon and the conclusion was:

- it should throw an error if you try to create a request with an amount much smaller than the ledger's supported scale
- you should be able to configure the rounding mode both on a per-receiver and per-request basis

(I'll implement this tomorrow)

sentientwaffle:
LGTM once npm-debug.log is gone

PR #45, opened 2016-09-13
Better support multiple receivers + multi-receivers

emschwartz:
Depends on https://github.com/interledger/five-bells-integration-test/pull/42
Should be merged after https://github.com/interledger/js-ilp/pull/42

Support multiple receivers listening on the same account by removing the behavior that would cause one receiver to cancel the incoming transfers destined for another.

Also support having a single receiver creating requests on behalf of multiple accounts by adding an optional account parameter to `createRequest`.

Fixes https://github.com/interledger/js-ilp/issues/43
Resolves https://github.com/interledger/js-ilp/issues/44


sentientwaffle:
LGTM

PR #46, opened 2016-09-13
feat: make payRequest idempotent

emschwartz:
uses a deterministic transfer uuid based on a sender secret and the condition to ensure that multiple calls to payRequest on the same sender with the same params will not submit two payments. a repeat payment will return the fulfillment of the original transfer.

resolves https://github.com/interledger/js-ilp/issues/40


sentientwaffle:
LGTM

PR #51, opened 2016-09-21
feat: add sender.stopListening

### LINKED ISSUE #49: Add sender.stopListening
To disconnect the client. Problem pointed out by @vhpoet

vhpoet:
Fixes #49

### DIFF
--- a/src/lib/sender.js
+++ b/src/lib/sender.js
@@ -191,11 +191,21 @@ function createSender (opts) {
       })
   }
 
+  /**
+   * Disconnect from the ledger and stop listening for events.
+   *
+   * @return {Promise.<null>} Resolves when the sender is disconnected.
+   */
+  function stopListening () {
+    return client.disconnect()
+  }
+
   return {
     quoteSourceAmount,
     quoteDestinationAmount,
     quoteRequest,
-    payRequest
+    payRequest,
+    stopListening
   }
 }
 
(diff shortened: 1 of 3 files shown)


emschwartz:
I think there's actually a better way to do this. The sender can automatically disconnect the client if it created it initially and there are no more listeners. If you pass in an `ilp-core` client when the sender is created, then it shouldn't automatically disconnect

src/lib/sender.js
```diff
@@ -188,11 +188,22 @@ function createSender (opts) {
       })
   }
 
+  /**
+   * Disconnect from the ledger and stop listening for events.
+   *
+   * @return {Promise.<null>} Resolves when the sender is disconnected.
+   */
+  function stopListening () {
+    client.removeListener('outgoing_fulfill', fulfillmentListener)
```
emschwartz: This will throw an error because `fulfillmentListener` is not defined in this scope.

Note each listener is removed [when the fulfillment is received](https://github.com/vhpoet/js-ilp/blob/02edb991fb259969196b1045013c693c59c62400/src/lib/sender.js#L181).

vhpoet: ah I already removed that, forgot to commit. Will do.

vhpoet:
@emschwartz done

emschwartz:
Cool, can you just run `npm run docs` and add the readme to update the docs?

vhpoet:
@emschwartz oh that was cool!

PR #64, opened 2017-01-17
Feat: add module for SPSP sending

sharafian:
Adds SPSP sending functionality to the ILP module. It uses a more functional style: instead of being a method on the sender, the SPSP module exposes functions that take a sender as the first argument. This avoids pointless wrapping of objects and events.

Quoting by source amount is done by quoting to get an initial destination amount, and then quoting again based on the amount that the `/receivers` endpoint returns in its payment request.

Not a breaking change. The sender and receiver are not changed at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,98 @@ co(function * () {
 
 ## API Reference
 
+<a name="module_SPSP..Client"></a>
+
+### SPSP~Client
+SPSP Client
+
+**Kind**: inner class of <code>[SPSP](#module_SPSP)</code>  
+
+* [~Client](#module_SPSP..Client)
+    * [new Client(opts)](#new_module_SPSP..Client_new)
+    * [.quoteSource](#module_SPSP..Client.Client+quoteSource) ⇒ <code>Promise.&lt;SPSPPayment&gt;</code>
+    * [.quoteDestination](#module_SPSP..Client.Client+quoteDestination) ⇒ <code>Promise.&lt;SPSPPayment&gt;</code>
+    * [.sendPayment](#module_SPSP..Client.Client+sendPayment) ⇒ <code>Promise.&lt;PaymentResult&gt;</code>
+    * [.query](#module_SPSP..Client.Client+query) ⇒ <code>Promise.&lt;Query&gt;</code>
+
+<a name="new_module_SPSP..Client_new"></a>
+
+#### new Client(opts)
+Create an SPSP client.
+
+
+| Param | Type | Description |
+| --- | --- | --- |
+| opts | <code>Object</code> | plugin options |
+| opts._plugin | <code>function</code> | (optional) plugin constructor. Defaults to PluginBells |
+
+<a name="module_SPSP..Client.Client+quoteSource"></a>
+
+#### client.quoteSource ⇒ <code>Promise.&lt;SPSPPayment&gt;</code>
+Get payment params via SPSP query and ILQP quote, based on source amount
+
+**Kind**: instance property of <code>[Client](#module_SPSP..Client)</code>  
+**Returns**: <code>Promise.&lt;SPSPPayment&gt;</code> - Resolves with the parameters that can be passed to sendPayment  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| receiver | <code>String</code> | webfinger identifier of receiver |
+| sourceAmount | <code>String</code> | Amount that you will send |
+
+<a name="module_SPSP..Client.Client+quoteDestination"></a>
+
+#### client.quoteDestination ⇒ <code>Promise.&lt;SPSPPayment&gt;</code>
+Get payment params via SPSP query and ILQP quote, based on destination amount
+
+**Kind**: instance property of <code>[Client](#module_SPSP..Client)</code>  
+**Returns**: <code>Promise.&lt;SPSPPayment&gt;</code> - Resolves with the parameters that can be passed to sendPayment  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| receiver | <code>String</code> | webfinger identifier of receiver |
+| destinationAmount | <code>String</code> | Amount that the receiver will get |
+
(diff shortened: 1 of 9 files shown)


emschwartz:
I'm not a fan of mixing the styles of the client and the more functional style that involves passing a client into another function. I think we either need to stick with wrapping the sender or completely switch to the more functional style (i.e. reimplementing ILQP in that style and calling those functions from the SPSP functions).

sharafian:
If we want to switch this repository to the functional style eventually, I don't think we should add new features in a style we're going to switch away from.

The SPSP functionality only adds two functions that don't require any state, so I don't think it makes sense to wrap every single function of the sender just for that.

README.md
```diff
@@ -152,6 +152,69 @@ co(function * () {
...
+
+#### client.quoteSource ⇒ <code>Promise.&lt;PaymentParams&gt;</code>
+Get payment params via SPSP request, based on source amount
+
+**Kind**: instance property of <code>[Client](#module_SPSP..Client)</code>  
+**Returns**: <code>Promise.&lt;PaymentParams&gt;</code> - Resolves with the parameters that can be passed to sendPayment  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| identifier | <code>String</code> | webfinger identifier of receiver |
```
emschwartz: I would call this just `receiver`

README.md
```diff
@@ -152,6 +152,69 @@ co(function * () {
...
+| sourceAmount | <code>String</code> | Amount that you will send |
+
+<a name="module_SPSP..Client.Client+quoteDestination"></a>
+
+#### client.quoteDestination ⇒ <code>Promise.&lt;PaymentParams&gt;</code>
+Get payment params via SPSP request, based on destination amount
```
emschwartz: Technically this is not "via SPSP request"

README.md
```diff
@@ -152,6 +152,69 @@ co(function * () {
...
+
+#### client.sendPayment ⇒ <code>Promise.&lt;PaymentResult&gt;</code>
+Sends a payment using the PaymentParams
+
+**Kind**: instance property of <code>[Client](#module_SPSP..Client)</code>  
+**Returns**: <code>Promise.&lt;PaymentResult&gt;</code> - Returns payment result  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| payment | <code>PaymentParams</code> | params, returned by quoteSource or quoteDestination |
```
emschwartz: `PaymentParams` should be documented

package.json
```diff
@@ -39,9 +39,11 @@
     "aguid": "^1.0.4",
     "bignumber.js": "^2.4.0",
     "canonical-json": "0.0.4",
+    "co-request": "^1.0.0",
```
emschwartz: I think we should use `superagent` instead of `co-request` because as far as I know it's a lighter dependency and we don't need all the extra functionality request includes

src/lib/spsp.js
```diff
@@ -0,0 +1,152 @@
...
+      */
+    this.sendPayment = sendPayment.bind(null, this.client)
+  }
+}
+
+module.exports = {
+  Client,
+  quoteSource,
+  quoteDestination,
+  sendPayment
```
emschwartz: Even though it's just an HTTP GET I think we should export a query function here. Maybe it would validate the schema or give you some error if that isn't a receiver

src/lib/spsp.js
```diff
@@ -0,0 +1,152 @@
...
+      _plugin: PluginBells
+    }, opts))
+
+    /**
+      * Get payment params via SPSP request, based on source amount
+      * @param {String} identifier webfinger identifier of receiver
+      * @param {String} sourceAmount Amount that you will send
+      * @returns {Promise.<PaymentParams>} Resolves with the parameters that can be passed to sendPayment
+      */
+    this.quoteSource = quoteSource.bind(null, this.client)
```
emschwartz: If you need to bind it to this client object it seems like it might make more sense to just make these normal methods on this object, instead of going for that functional style. I think that style only makes sense if we're passing in the plugin directly

sharafian:
@emschwartz I've addressed your changes on this PR. The functions now take plugins instead of ILP Core clients, and there is a section in the README that documents the `SPSPPayment` object type returned by the SPSP quote methods. Query is also exposed by the module now.

sharafian:
Now uses SPSP spec from https://gist.github.com/sharafian/df7a4b7e2ff000248800b113f06f549a , with some modifications to address the comments.

src/lib/spsp.js
```diff
@@ -0,0 +1,189 @@
...
+      ']')
+  }
+
+  return quote
+}
+
+const _createPayment = (spsp, quote) => {
+  return {
+    id: uuid(),
+    source_amount: quote.sourceAmount,
```
emschwartz: I think we should use camelCase here (this doesn't match the documented fields)

test/spspSpec.js
```diff
@@ -0,0 +1,194 @@
...
+    it('should fail without an identifier', function * () {
+      yield expect(SPSP.quoteSource(this.plugin, undefined, '10')).to.eventually.be.rejected
+    })
+
+    it('should fail without an amount', function * () {
+      yield expect(SPSP.quoteSource(this.plugin, 'alice@example.com')).to.eventually.be.rejected
+    })
+  })
+
+  describe('sendPayment', function () {
```
emschwartz: Need a test to make sure it works

test/spspSpec.js
```diff
@@ -0,0 +1,194 @@
...
+const spspResponse = {
+  shared_secret: 'itsasecret',
+  destination_account: 'example.alice',
+  maximum_destination_amount: '20',
+  minimum_destination_amount: '10'
+}
+
+const SPSP = require('../src/lib/spsp')
+
+describe('SPSP Module', function () {
```
emschwartz: query could use a test as well

PR #66, opened 2017-01-21
fix: ignore packets created by other receivers

emschwartz:
the request-specific address now includes a receiverId (deterministically generated from the hmacKey) so that a given receiver will not reject transfers requested by another application listening on the same account


src/lib/receiver.js
```diff
@@ -302,17 +319,25 @@ function createReceiver (opts) {
   })
 }
 
+// "~ipr" + base64url encoded 8-byte id
+function generateReceiverId (hmacKey) {
+  const hmac = crypto.createHmac('sha256', hmacKey)
+  hmac.update(RECEIVER_ID_PREFIX, 'utf8')
```
justmoon: Why is `RECEIVER_ID_PREFIX` used here? I would have expected a separate constant using a string like `"receiver_id"`. Also, all the constants that are used with a given HMAC key should be tracked in one place. If there is any ambiguity, it can be a very serious vulnerability.

Another option would be to have a class that wraps the HMAC key and keeps it as a private field. That way you can easily audit that class cryptographically, without having to trace everywhere the HMAC key is being passed around.

justmoon:
Rebased to fix tests.

test/data/transferIncoming.json
```diff
@@ -5,7 +5,7 @@
   "amount": "1.0000",
   "data":{
     "ilp_header": {
-      "account": "ilpdemo.blue.bob.~ipr.D6-mqxYanek.22e315dc-3f99-4f89-9914-1987ceaa906d",
+      "account": "ilpdemo.blue.bob.~ipr.65f8947458feb55c.22e315dc-3f99-4f89-9914-1987ceaa906d",
```
justmoon: I think base64url is fine here. Let's be clear that ILP addresses are case-sensitive. Having a widely used module like this require case-sensitivity early on, it may actually help ensure it's consistently enforced.

src/lib/receiver.js
```diff
@@ -185,6 +193,18 @@ function createReceiver (opts) {
       return rejectIncomingTransfer(transfer.id, 'no-packet')
     }
 
+    // check if the address ends with "~ipr" + our receiver id +
+    const myRequestAddressRegex = new RegExp(receiverId + '\\..+$')
```
sentientwaffle: `receiverId` here includes `RECEIVER_ID_PREFIX` (`'~ipr.'`), so the `'.'` will be treated as a wildcard. Does that matter?

justmoon: Good catch! That would matter. Because it requires the next character after the wildcard character to be a period `.` - which it won't be if the period was already matched by the wildcard. So yeah, I don't think this code works, unless I'm missing something.

PR #67, opened 2017-01-30
Add PSK functionality

justmoon:


README.md
```diff
@@ -150,6 +150,56 @@ co(function * () {
...
+  yield receiver.listen()
+  receiver.on('incoming', (transfer, fulfillment) => {
+    console.log('received transfer:', transfer)
+    console.log('fulfilled transfer hold with fulfillment:', fulfillment)
+  })
+
+  const secret = receiver.generateSharedSecret()
+  console.log('secret:', secret)
+
+  const request = sender.createRequest(Object.assign({}, secret, {
```
emschwartz: Huh? The sender creates requests?

README.md
```diff
@@ -150,6 +150,56 @@ co(function * () {
...
+
+const receiver = ILP.createReceiver({
+  _plugin: FiveBellsLedgerPlugin,
+  account: 'https://localhost/ledger/accounts/bob',
+  password: 'bobbob'
+})
+
+co(function * () {
+  yield receiver.listen()
+  receiver.on('incoming', (transfer, fulfillment) => {
```
emschwartz: Is the receiver going to fulfill everything that comes in? Wouldn't you want a callback function or something to let the application check whether they want the money?

src/lib/receiver.js
```diff
@@ -1,15 +1,17 @@
...
 const EventEmitter = require('eventemitter2')
+const base64url = require('../utils/base64url')
+const cc = require('../utils/condition')
 const debug = require('debug')('ilp:receiver')
 const BigNumber = require('bignumber.js')
 const createHmacHelper = require('../utils/hmac').createHmacHelper
 
-const RECEIVER_ID_PREFIX = '~ipr.'
+const IPR_RECEIVER_ID_PREFIX = '~ipr.'
+const SSP_RECEIVER_ID_PREFIX = '~ssp.'
```
emschwartz: I think it would be better to separate the IPR and SSP receivers into completely separate modules. Chances are you won't be using both at the same time and I think it could cause issues having the same receiver instance listening on two separate protocols.

src/lib/receiver.js
```diff
@@ -186,16 +203,31 @@ function createReceiver (opts) {
...
     if (!packet) {
       debug('got notification of transfer with no packet attached')
       return rejectIncomingTransfer(transfer.id, 'no-packet')
     }
 
-    // check if the address contains "~ipr" + our receiver id
-    if (packet.account.indexOf(receiverId) === -1) {
-      debug('got notification of transfer for another receiver: ' + packet.account + ' my id is: ' + receiverId)
+    // check if the address starts with our address
+    if (packet.account.indexOf(getAddress()) !== 0) {
```
emschwartz: The reason this looks for the receiverId instead of the address is because right now you can pass the address to the `createRequest` function. That was intended for use cases where you want to have one receiver creating and fulfilling requests on behalf of multiple accounts, like in the ilp-kit, but that isn't currently being used. Right now the `ilp-kit` is making one new receiver per payment, but I'm not sure that's the best way to set this up. If we want to keep that behavior we can take out the ability to specify the account for each request.

src/lib/sender.js
```diff
@@ -189,6 +196,40 @@ function createSender (opts) {
...
   /**
+   * Create a payment request using an SSP shared secret.
+   *
+   * @param {Object} params Parameters for creating payment request
+   * @param {String} params.destination_amount Amount that should arrive in the recipient's account
+   * @param {String} params.destination_account Target account's ILP address
+   * @param {String} [params.id=uuid.v4()] Unique ID for the request (used to ensure conditions are unique per request)
+   * @param {String} [params.expires_at=30 seconds from now] Expiry of request
+   * @param {Object} [params.data=null] Additional data to include in the request
+   *
```
sentientwaffle: Should `shared_secret` be mentioned as a parameter here?

src/lib/sender.js
```diff
@@ -189,6 +196,40 @@ function createSender (opts) {
...
+      expires_at: params.expires_at || moment().add(defaultRequestTimeout, 'seconds').toISOString()
+    }
+
+    if (params.data) {
+      paymentRequest.data = params.data
+    }
+
+    paymentRequest.address += '.' + (params.id || uuid.v4())
+
+    const sharedSecret = Buffer.from(params.shared_secret, 'base64')
```
sharafian: `params.shared_secret` is missing in the JSdoc for this function

src/lib/sender.js
```diff
@@ -189,6 +196,40 @@ function createSender (opts) {
...
+    const conditionPreimage = hmacHelper.hmacJsonForSspCondition(paymentRequest, sharedSecret)
+    const condition = toConditionUri(conditionPreimage)
```
sharafian: why does this use a different `toConditionUri` function from the receiver?

src/utils/hmac.js
```diff
@@ -17,13 +19,32 @@ function createHmacHelper (hmacKey) {
...
+  function hmacJsonForSspCondition (obj, sharedSecret) {
+    const sspConditionKey = hmac(sharedSecret, SSP_CONDITION_STRING)
+    const jsonString = stringify(obj)
+    const hmacDigest = hmac(sspConditionKey, jsonString)
+    return hmacDigest
+  }
+
   return {
     hmacJsonForIprCondition,
-    getIprReceiverId
```
sharafian: How come IPR receiver has been changed to just receiver in the code? Wouldn't it make sense to be more specific now that we have two protocols being used in here?

justmoon: There is no reason the receiver ID wouldn't be shared between IPR and SSP.

src/lib/sender.js
justmoon: Hmm? It is using the same one!

src/lib/receiver.js
```diff
@@ -256,15 +295,25 @@ function createReceiver (opts) {
...
+        if (protocol === 'ipr') {
+          /**
+           * [IncomingTransfer](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#incomingtransfer) from the ledger plugin and the fulfillment string for a specific request
+           *
+           * @event incoming:<requestid>
+           * @type {object}
+           */
+          // Allow listeners for specific requests and on wildcard events such that
+          // `incoming:appid.*` will match `incoming:appid:requestid`
+          eventEmitter.emit('incoming:' + requestId, transfer, fulfillment)
```
justmoon: Because it would be a breaking change.

src/lib/receiver.js
justmoon: Hmm, interesting. Right now it seems that there is 1:1:1 relationship between receiver, client and plugin. (See `getAddress()`.)

If we switch to a multi-plugin receiver, we would still know which plugin emitted the event, so we'd still know what our address is supposed to be. So we could check it. I can't think of a strong argument for or against verifying the address. Maybe someone else has one?

README.md
justmoon: Yeah, the reason I implemented it this way is because I didn't want to create a second quoting method.

You can think of it like this: In SSP, both the sender and the receiver can create requests (because the secret is shared). (So the sender doesn't have to ask the receiver for a request, it can just generate one.) IPR is basically the same as SSP, except the receiver does not share the secret, so only the receiver can generate requests.

Not sure if "request" is the right name, it's more of a "payment description" or something. You called it "packet", but I don't think it's the same as an ILP packet, so I'd prefer calling it something else.

README.md
justmoon: Good idea, I'll think about how to add something like that!

src/lib/receiver.js
justmoon: As per another comment - they both hash the "request"/"packet" the same way - so we could split them into separate modules, but there would be a lot of overlap. You can kind of tell from this PR - there are only a handful of lines actually dedicated to only IPR or only SSP, 95% of the sender and receiver code is the same for both.

So we should chat about it - but it seems like having two different receivers seems like we're overengineering it a bit.

> Chances are you won't be using both at the same time and I think it could cause issues having the same receiver instance listening on two separate protocols.

I don't think so - at least if we make it so that SSP requires a callback to be registered with the receiver. If you only want to use IPR, you're fine, because there is no callback registered, so this receiver will reject SSP payments. And if you only want to use SSP, you're fine because you haven't issued any payment requests, so you won't get any valid IPR payments.

(The latter is a bit sketchier, because it relies on cryptography. Personally, I think it would be fine. But if we think it isn't, we could just add a method to enable IPR.)

src/lib/sender.js
sharafian: The receiver uses `cc.toConditionUri` and the sender uses `toConditionUri` from `util/condition`

justmoon: They're the same thing! ^^

`cc` == `util/condition`. The only reason it's called `cc` in the receiver is because it uses both `toConditionUri` and `toFulfillmentUri` from `util/condition`. :)

src/lib/receiver.js
```diff
@@ -168,13 +185,13 @@ function createReceiver (opts) {
   /**
    * @private
    *
-   * When we receive transfer notifications, check the transfers
-   * and try to fulfill the conditions (which will only work if
-   * they correspond to requests we created)
+   * When we receive a transfer notification, check the transfer
+   * and try to fulfill the condition (which will only work if
+   * it corresponds to a request we created)
```
emschwartz: The last part of this comment is not true in the case of SSP

src/lib/receiver.js
emschwartz: Might not be bad to add the `incoming:ipr` event as well

src/lib/sender.js
sharafian: shouldn't this shared secret be parsed from base64url? `generateSharedSecret` returns a shared secret encoded in base64url.

emschwartz: Also, it might be useful to allow a Buffer to be passed in. If you're sending multiple payments there's no reason to do the base64 decoding every time

justmoon: `Buffer.from(..., 'base64')` supports both base64 and base64url inputs.

justmoon: This code will work fine for that. If `shared_secret` is a buffer, `Buffer.from` just becomes a no-op.

justmoon:
Should we add AES-CTR (256 bit) support before merging? I vote yes.

@emschwartz @sharafian

emschwartz:
Works for me.

Also, merge https://github.com/interledgerjs/ilp/pull/68 before merging this PR

sharafian:
@justmoon implemented AES encryption of the payment request data in https://github.com/interledgerjs/ilp/pull/68#discussion_r100033569

emschwartz:
After trying out the name KEP for a couple of days, I have to say I'm not a fan. If you pronounce it normally (not like "keep") it sounds silly, and if you pronounce it like "keep" it sounds more like a defunct hard drive encryption startup than a protocol.

Since the only alternative that's been proposed is SSP, I'm going to say I think we should stick with that. I'll admit that the similarity between SPSP and SSP is confusing, but at least one uses the other now. That seemed really bad when SPSP was not built on SSP, but I think it's acceptable now.

emschwartz:
Discussed with @sharafian and @michielbdejong and we all liked calling it PSK for Pre-Shared Key. https://github.com/interledger/rfcs/issues/43#issuecomment-278616337

README.md
emschwartz: @sharafian can you update the example with the listen callback?

README.md
emschwartz: @justmoon @sharafian should we change the name from request to `PaymentParameters` before merging this?

README.md
```diff
@@ -150,6 +150,56 @@ co(function * () {
 
 ```
 
+### Shared Secret Example
```
emschwartz: We should also update the description of this module at the top of the README to talk about PSK, how to use that, and how it's different from IPR

src/lib/receiver.js
emschwartz: This comment should be changed

src/lib/receiver.js
emschwartz: I'm fine with checking the account. If we make this change though we should remove the option to specify the account in the `createRequest` call (and note that that is technically a breaking change)

src/lib/receiver.js
```diff
@@ -211,7 +255,22 @@ function createReceiver (opts) {
...
 
     if (packet.data && packet.data.data) {
-      paymentRequest.data = packet.data.data
+      if (protocol === 'psk') {
+        if (Object.keys(packet.data.data).length > 1 || !packet.data.data.blob) {
+          return Promise.reject('payment request data should only include encrypted blob. got: ' +
+            JSON.stringify(packet.data.data))
+        }
+        try {
+          paymentRequest.data = cryptoHelper.aesDecryptObject(
```
emschwartz: Is there a size limit on the encrypted blob?

src/lib/sender.js
emschwartz: This needs to be added

test/receiverSpec.js
```diff
@@ -518,6 +519,103 @@ describe('Receiver Module', function () {
           const results = yield this.client.emitAsync('incoming_prepare', this.transfer)
           expect(spy).not.to.have.been.called
         })
+
+        it('should reject PSK payment without reviewPayment', function * () {
```
emschwartz: I think it would be good to have a test for a PSK packet that has been tampered with (i.e. the condition doesn't match)

src/lib/receiver.js
sharafian: the size of the blob should be restricted by the size of the data field

emschwartz: ok, I guess we'll wait until that's specified

emschwartz:
We should change what the HMAC is computed on from `canonical-json` to the [format](https://gist.github.com/sharafian/df7a4b7e2ff000248800b113f06f549a#gistcomment-1986027) suggested by @justmoon before this is merged, because that will be a breaking change. Note it can be updated for IPR as well because for that one only the receiver cares how the fulfillment is generated.

Also, https://github.com/interledgerjs/ilp/pull/70 fixes most of the things I was thinking of commenting on in a review.

docs/README.template.md
```diff
@@ -25,106 +25,127 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
...
+### IPR Sending and Receiving Example
 
 ```js
 'use strict'
 
+const co = require('co')
 const ILP = require('ilp')
 const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
+
+const sender = ILP.createSender(i
```
sharafian: typo

PR #68, opened 2017-02-07
Feat: add callback to listen

sharafian:
Adds a callback to `listen`, which is triggered right before the receiver fulfills an incoming payment. The callback is called with the payment request object, and will stop the receiver from fulfilling if it throws an error.


README.md
```diff
@@ -357,14 +362,19 @@ Generate shared secret for Shared Secret Protocol (SSP).
...
 conditions for transfers corresponding to requests this
 receiver created.
 
 **Kind**: inner method of <code>[createReceiver](#module_Receiver..createReceiver)</code>  
 **Returns**: <code>Promise.&lt;null&gt;</code> - Resolves when the receiver is connected  
 **Emits**: <code>[incoming](#event_incoming)</code>, <code>incoming:&lt;requestid&gt;</code>, <code>incoming:ssp:&lt;token&gt;</code>  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| callback | <code>function</code> | (optional) called when an incoming payment is about to be fulfilled. throwing an error will stop from fulfilling. |
```
emschwartz: Not a fan of depending on a callback throwing an error. That's unusual. Either it should take a function that gets the request and a callback or it should take a function that returns a promise.

sentientwaffle:
Why is this set up as an parameter of `receiver.listen()` instead of adding an option to `createReceiver()`? Also, a test for this behavior would be good.

sharafian:
@emschwartz @sentientwaffle addressed your comments

src/lib/receiver.js
```diff
@@ -281,9 +289,13 @@ function createReceiver (opts) {
     }
 
     const fulfillment = cc.toFulfillmentUri(conditionPreimage)
-    debug('about to submit fulfillment: ' + fulfillment)
+
     // returning the promise is only so the result is picked up by the tests' emitAsync
-    return client.fulfillCondition(transfer.id, fulfillment)
+    return reviewPayment(paymentRequest)
```
sentientwaffle: If `reviewPayment()` rejects, the `catch` will print `'error submitting fulfillment', err`, which isn't accurate in this case.

sharafian: :+1: Fixed so that it rejects to the `reviewPayment`'s error if `reviewPayment` rejects.

src/lib/receiver.js
```diff
@@ -18,6 +18,12 @@ const SSP_RECEIVER_ID_PREFIX = '~ssp.'
  */
 
 /**
+ * @callback reviewPaymentCallback
```
emschwartz: I think it's strange to say a callback is returning a promise

src/lib/receiver.js
```diff
@@ -49,6 +56,7 @@ function createReceiver (opts) {
   const sspReceiverId = SSP_RECEIVER_ID_PREFIX + receiverId + '.'
   debug('receiver id: ' + receiverId)
 
+  const reviewPayment = opts.reviewPayment || (() => Promise.resolve(null))
```
emschwartz: I'm not sure it's a good idea to make the default behavior "accept any money that comes in"

sharafian: That's required unless we want to make this a breaking change

src/lib/receiver.js
sharafian: Is there a different way to express a function that returns a promise using JSdoc?

src/lib/receiver.js
emschwartz: This is one reason having a receiver listening for SSP/KEP and IPR payments at the same time is problematic. I think the review payment function should be required to enable SSP/KEP payments, and we could do that without making a breaking change

src/lib/receiver.js
emschwartz: No, you're right. That's fine

src/lib/receiver.js
sharafian: Additionally, most users won't have any backend to determine whether to accept payments, and so they'd end up writing a function like this most of the time anyways.

emschwartz: In general it's a bad idea to default to "accept money from anyone, no matter what". Instead of using the presence of this callback to determine whether or not to enable SSP/KEP (we need to decide on the name), we could also have an explicit option to enable that other protocol that needs a name. Not sure that's better. However, adding this new protocol to the `ilp` receiver does make the very big difference that money can show up in your account without you expecting or wanting it, which is not possible with pure IPR.

test/receiverSpec.js
```diff
@@ -478,6 +483,19 @@ describe('Receiver Module', function () {
           expect(results).to.deep.equal(['sent'])
         })
 
+        it('should not fulfill payment if reviewPayment callback rejects', function * () {
```
emschwartz: This needs a test to make sure that if you instantiate what you think is just an IPR receiver you can't accidentally get money showing up in your account with KEP

src/lib/receiver.js
justmoon: Agree with Evan - you should have to turn on SSP explicitly by providing a `reviewPayment` function.

I think it would be acceptable if `reviewPayment` was allowed to return either `null` or `Promise<null>`.

That way, I can do `reviewPayment: () => null` and just generally skip having to generate a Promise if I'm only going to do synchronous checks.

src/lib/receiver.js
```diff
@@ -281,44 +289,48 @@ function createReceiver (opts) {
     }
 
     const fulfillment = cc.toFulfillmentUri(conditionPreimage)
-    debug('about to submit fulfillment: ' + fulfillment)
+
     // returning the promise is only so the result is picked up by the tests' emitAsync
-    return client.fulfillCondition(transfer.id, fulfillment)
+    return reviewPayment(transfer)
```
justmoon: You should catch sync errors and treat them the same way as async errors. I'd just do:

``` js
Promise.resolve(transfer)
  .then(reviewPayment)
  // same code after that
```

src/lib/receiver.js
```diff
@@ -281,44 +289,48 @@ function createReceiver (opts) {
...
+            if (protocol === 'ipr') {
+              /**
+              * [IncomingTransfer](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#incomingtransfer) from the ledger plugin and the fulfillment string for a specific request
+              *
+              * @event incoming:<requestid>
+              * @type {object}
+              */
+              // Allow listeners for specific requests and on wildcard events such that
+              // `incoming:appid.*` will match `incoming:appid:requestid`
+              eventEmitter.emit('incoming:' + requestId, transfer, fulfillment)
```
justmoon: Looking at this now... I doubt we'll break a lot of code if we change this to `incoming:ipr:`, so maybe we should just rip off the band-aid. :)

README.md
```diff
@@ -389,4 +395,13 @@ Disconnect from the ledger and stop listening for events.
...
+**Kind**: event emitted by <code>[Receiver](#module_Receiver)</code>  
+<a name="module_Receiver..reviewPaymentCallback"></a>
+
+### Receiver~reviewPaymentCallback ⇒ <code>Promise.&lt;null&gt;</code>
+**Kind**: inner typedef of <code>[Receiver](#module_Receiver)</code>  
+**Returns**: <code>Promise.&lt;null&gt;</code> - cancels the payment if it rejects.  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| request | <code>PaymentRequest</code> | payment request object for the payment being reviewed |
```
justmoon: I'm confused - this says that the `reviewPayment` callback gets the payment request, but in the code it seems to get the transfer.

I'd probably pass `(request, transfer)` with the rationale that 90% of the time you don't want to parse the transfer yourself, but sometimes you may want to dig deeper.

We should think about a different name for the payment request. That name is great for IPR, but since it's the exact same string/object between KEP and IPR, it should probably have a name that works for both?

emschwartz: Payment summary, body, details -- or just "Payment"?

src/lib/receiver.js
emschwartz: Why not have both and deprecate the old one?

README.md
sharafian: payment parameters?

src/utils/hmac.js
```diff
@@ -39,12 +40,42 @@ function createHmacHelper (hmacKey) {
     return hmacDigest
   }
 
+  // turn object into encrypted buffer
+  function aesEncryptObject (obj, sharedSecret) {
```
emschwartz: Arguably this stuff shouldn't be in an "hmac" class

test/receiverSpec.js
```diff
@@ -529,18 +529,43 @@ describe('Receiver Module', function () {
...
             destination_amount: "1",
+            data: { for: 'that thing' }
           }))
 
           this.transfer.account = request.address
           this.transfer.executionCondition = request.condition
           this.transfer.data.ilp_header.data.expires_at = request.expires_at
           this.transfer.data.ilp_header.account = request.address
-          delete this.transfer.data.ilp_header.data.data
+          this.transfer.data.ilp_header.data.data = request.data
```
emschwartz: Can you add a test specifically for the fact that the sender automatically encrypts the data, and that the receiver expects it to be encrypted? Also, what would happen if the data is not encrypted? Would the receiver understand it?

test/senderSpec.js
```diff
@@ -28,6 +30,51 @@ describe('Sender Module', function () {
...
+      const request = this.sender.createRequest(Object.assign({}, this.kep, {
+        destination_amount: '1'
+      }))
+
+      assert.match(request.address, new RegExp('^' + this.kep.destination_account))
+      assert.equal(request.amount, '1')
+      assert.equal(request.expires_at, '1970-01-01T00:00:30.000Z')
+    })
+
+    it('should encrypt the payment request data and store as a base64-encoded blob', function () {
```
emschwartz: I'm not sure this makes sense. If you encrypt data that you're giving to the sender, just for them to send it back to you, what's the point of putting it there at all?

test/receiverSpec.js
```diff
@@ -529,18 +529,43 @@ describe('Receiver Module', function () {
...
+          const request = sender.createRequest(Object.assign({}, kep, {
+            destination_amount: "1",
+            data: { for: 'that thing' }
+          }))
+
+          this.transfer.account = request.address
+          this.transfer.executionCondition = request.condition
+          this.transfer.data.ilp_header.data.expires_at = request.expires_at
+          this.transfer.data.ilp_header.account = request.address
+          this.transfer.data.ilp_header.data.data = {
```
emschwartz: `transfer.data.ilp_header.data.data.blob` is absurd. Can we please switch to the new packet format?? @justmoon

justmoon: I'm basically happy with the format in the gist. Let's give everyone a last chance for feedback, then make the switch. Good?

Btw: The new format won't totally fix the data.data.data thing. That's an artifact of a layered approach. If `data` means "next layer" and we want to keep that consistent...

README.md
```diff
@@ -281,14 +281,19 @@ Pay for a payment request. Uses a determinstic transfer id so that paying is ide
 <a name="module_Sender..createSender..createRequest"></a>
 
 #### createSender~createRequest(params) ⇒ <code>Object</code>
-Create a payment request using an SSP shared secret.
+Create a payment request using a PSK shared secret.
```
emschwartz: Hehe, "PSK shared secret" is redundant in the same way "IP Protocol" is :)

src/lib/receiver.js
```diff
@@ -30,6 +38,7 @@ const SSP_RECEIVER_ID_PREFIX = '~ssp.'
  * @param  {Boolean} [opts.allowOverPayment=false] Allow transfers where the amount is greater than requested
  * @param {String} [opts.roundingMode=null] Round request amounts with too many decimal places, possible values are "UP", "DOWN", "HALF_UP", "HALF_DOWN" as described in https://mikemcl.github.io/bignumber.js/#constructor-properties
  * @param  {Number} [opts.connectionTimeout=10] Time in seconds to wait for the ledger to connect
+ * @param  {reviewPaymentCallback} [opts.reviewPayment] called before fulfilling any incoming payments. The receiver doesn't fulfill the payment if reviewPayment rejects.
```
emschwartz: If this is required to enable PSK payments, the docs should say so

src/lib/receiver.js
```diff
@@ -263,14 +290,18 @@ function createReceiver (opts) {
       return rejectIncomingTransfer(transfer.id, 'expired')
     }
 
+    if (protocol === 'psk' && !reviewPayment) {
+      debug('got PSK payment on non-PSK receiver')
+      return Promise.reject(new Error('opts.reviewPayment was not specified, so PSK is not supported on this receiver'))
```
emschwartz: I think this should call `return rejectIncomingTransfer(transfer.id, 'psk-not-supported')` instead

src/lib/receiver.js
```diff
@@ -263,14 +290,18 @@ function createReceiver (opts) {
...
     if (protocol === 'ipr') {
       conditionPreimage = hmacHelper.hmacJsonForIprCondition(paymentRequest)
-    } else if (protocol === 'ssp') {
-      const sharedSecret = hmacHelper.getSspSharedSecret(requestId)
-      conditionPreimage = hmacHelper.hmacJsonForSspCondition(paymentRequest, sharedSecret)
+    } else if (protocol === 'psk') {
+      conditionPreimage = cryptoHelper.hmacJsonForPskCondition(paymentRequest, sharedSecret)
     } else {
-      throw new Error('Invalid protocol')
+      return Promise.reject(new Error('Invalid protocol'))
```
emschwartz: This should also reject the transfer

test/receiverSpec.js
```diff
@@ -518,6 +519,102 @@ describe('Receiver Module', function () {
...
+            amount: 1,
+            id: '22e315dc-3f99-4f89-9914-1987ceaa906d',
+            expiresAt: this.transfer.data.ilp_header.data.expires_at,
+            data: { for: 'that thing' }
+          })
+
+          this.transfer.executionCondition = request.condition
+          yield expect(client.emitAsync('incoming_prepare', this.transfer))
+            .to.be.rejectedWith('rejected!')
+        })
```
emschwartz: How about adding another test to make sure it rejects payments that have the right receiver ID but an unknown protocol?

test/receiverSpec.js
```diff
@@ -518,6 +519,102 @@ describe('Receiver Module', function () {
           const results = yield this.client.emitAsync('incoming_prepare', this.transfer)
           expect(spy).not.to.have.been.called
         })
+
+        it('should not fulfill PSK payment without reviewPayment', function * () {
```
emschwartz: It should actually reject the transfer

test/receiverSpec.js
sharafian: There's not any execution path that can trigger that behavior, currently. Either protocol is set to `ipr`, `psk`, or a `"not-my-packet"` rejection is returned.

src/lib/receiver.js
```diff
@@ -263,14 +290,18 @@ function createReceiver (opts) {
...
     if (protocol === 'ipr') {
       conditionPreimage = hmacHelper.hmacJsonForIprCondition(paymentRequest)
-    } else if (protocol === 'ssp') {
-      const sharedSecret = hmacHelper.getSspSharedSecret(requestId)
-      conditionPreimage = hmacHelper.hmacJsonForSspCondition(paymentRequest, sharedSecret)
+    } else if (protocol === 'psk') {
+      conditionPreimage = cryptoHelper.hmacJsonForPskCondition(paymentRequest, sharedSecret)
     } else {
-      throw new Error('Invalid protocol')
+      return rejectIncomingTransfer(transfer.id, 'invalid-protocol')
```
emschwartz: @sharafian if it's impossible to get here, we should take out this branch

sharafian: If the code changes in the future, don't we want this to catch possible bugs?

emschwartz: Not a huge fan of having untestable lines of code in there

README.md
```diff
@@ -357,7 +357,7 @@ Create a payment request
 <a name="module_Receiver..createReceiver..generateSharedSecret"></a>
 
 #### createReceiver~generateSharedSecret() ⇒ <code>Object</code>
-Generate shared secret for Shared Secret Protocol (KEP).
+Generate shared secret for Shared Secret Protocol (PSK).
```
justmoon: The "Shared Secret Protocol", or "PSK" for short. ^^

PR #69, opened 2017-02-09
fix: race condition when receiving fulfillment

justmoon:
Transfers may be fulfilled between when the sender is checking if they are fulfilled and when the sender starts listening to `outgoing_fulfilled` events.


vhpoet:
we need to write tests for tests

PR #70, opened 2017-02-14
Fix and test PSK, simplify docs, remove inconsistent snake_case

emschwartz:


src/lib/receiver.js
```diff
@@ -170,13 +176,13 @@ function createReceiver (opts) {
...
    *
-   * @return {Object} Object containing destination address and shared secret
+   * @return {PSKParams}
    */
   function generateSharedSecret () {
     const token = base64url(hmacHelper.getPskToken())
     return {
-      destination_account: getAddress() + '.' + pskReceiverId + token,
-      shared_secret: base64url(hmacHelper.getPskSharedSecret(token))
+      destinationAccount: getAddress() + '.' + pskReceiverId + token,
```
justmoon: Hmm, the snake case here wasn't random, I was thinking that you might use these values in other applications, so the field names should be standardized. It's obviously not a big deal, it's easy enough to change it manually if another implementation expects `destination_account`.

We decided to use snake case for all protocols. The question is whether JSONifying these values and copying them into a config file somewhere constitutes a "protocol".

src/lib/receiver.js
```diff
@@ -26,6 +26,12 @@ const PSK_RECEIVER_ID_PREFIX = '~psk.'
  */
 
 /**
+ * @typedef {Object} PSKParams
```
justmoon: Should probably be `PskParams` - not saying the code is currently consistent, but I'm seeing more HttpRequest than HTTPRequest.

src/lib/receiver.js
emschwartz: For now I don't think this return value constitutes a protocol by itself so I think we should go with camelCase

src/lib/receiver.js
emschwartz: Fixed

README.md
```diff
@@ -114,7 +59,7 @@ const co = require('co')
 const ILP = require('ilp')
 const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
 
-const sender = ILP.createSender({
+const sender = ILP.createSender(i
```
sharafian: Typo

Issue #63, opened 2016-12-29
Refactor `ilp` module


emschwartz:
(with @sharafian)

The ILP module should have the following exports:
- `SPSP` - SPSP client (and server?)
- `ILQP` - implements Interledger Quoting Protocol (refactored out of [`ilp-core`](https://github.com/interledgerjs/ilp-core/issues/76))
- `IPR` - implements the Interledger Payment Request (with [this update](https://github.com/interledger/rfcs/issues/138))
- `PluginCommon` - when we switch to the [Common Ledger API](https://github.com/interledger/rfcs/pull/125), we should bundle that plugin with this module and use it by default (we might want to bundle `ilp-plugin-bells` in the meantime)
- `Packet`? - it might be useful to bundle some helper functions for creating ILP packets, though this might not be necessary as long as we stick with a JSON packet

And it could export a couple of helper functions (from the root `ilp`) that bundle common steps using the aforementioned modules:
- `constructTransferFromSource` - this would return a [Transfer](https://github.com/interledger/rfcs/issues/139) object with an ILP packet attached. If the account identifiers given are ILP addresses, it would just use `ILQP` for quoting and construct the transfer from there. If they are SPSP identifiers, it could detect that and use `SPSP` to resolve those to ILP addresses
- `constructTransferFromDestination` - same as above but based on a fixed destination amount
- `constructTransferFromPacket` - basically quoting a [Payment Request](https://github.com/interledger/rfcs/issues/138)
- `sendTransfer` - the main difference between this and a plugin's `sendTransfer` function would be that this one would only resolve once it gets the fulfillment (as opposed to just when the transfer is accepted by the ledger)

As suggested by @sharafian, we can get rid of the unusual quote format from `ilp-core` and replace it with normal Transfer objects (that include ILP packets). If we want to provide easier access to fields like the destination amount that are nested in the ILP packet, we can add some extra helper functions for that.

PR #79, opened 2017-03-08
Feat: [BREAKING] rewrite module in a functional style

sharafian:
Instead of having many different state-ful client and receiver classes, all state is kept to the plugin. Individual functions can be used for quoting, parsing the ILP packet (currently the JSON version), etc.

The SPSP module provides a little more helpful defaults, generating an "SPSPPayment" object from a quote, and then sending that with the `sendPayment` function.

The `listen` function for receivers now returns a `stopListening` function, and takes the "reviewPayment" callback as a parameter after plugin and options. The callback is given the payment details and a fulfill function, allowing it to use its own async or synchronous logic to fulfill the transfer or not.

Depends on https://github.com/interledger/rfcs/pull/166
Depends on https://github.com/interledger/rfcs/pull/169
Depends on https://github.com/interledger/rfcs/pull/170
Resolves https://github.com/interledgerjs/ilp/issues/63


test/packetSpec.js
```diff
@@ -0,0 +1,123 @@
...
+    })
+
+    it('should parse a packet w/o data', function () {
+      delete this.packet.ilp_header.data
+      Packet.parse(this.packet)
+    })
+
+    it('should parse a packet w/o data.data', function () {
+      delete this.packet.ilp_header.data.data
+      Packet.parse(this.packet)
```
emschwartz: This should test what it actually returns

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+        response,
+        this.result)
+    })
+
+    it('should default to getInfo\'s connectors', function * () {
+      // remove manually provided connectors
+      delete this.params.connectors
+
+      this.result.expiresAt = moment().format()
+      const response = yield ILQP.quote(this.plugin, this.params)
```
emschwartz: How does this actually test that it uses the `getInfo` connectors?

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+    })
+
+    it('should throw if source and dest amounts are defined', function * () {
+      this.params.destinationAmount = this.params.sourceAmount = '1'
+
+      yield expect(ILQP.quote(this.plugin, this.params))
+        .to.be.rejectedWith(/provide source or destination amount but not both/)
+    })
+
+    it('should return undefined if there are no connectors', function * () {
```
emschwartz: It should throw an error instead. Returning undefined is pretty unhelpful

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+      this.params.connectors = []
+      const response = yield ILQP.quote(this.plugin, this.params)
+      assert.equal(response, undefined)
+    })
+
+    it('should return undefined on a timeout', function * () {
```
emschwartz: Should throw

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+      this.plugin.sendMessage = () => Promise.resolve(null)
+      this.params.timeout = 10
+      const response = yield ILQP.quote(this.plugin, this.params)
+      assert.equal(response, undefined)
+    })
```
emschwartz: What happens if the connector responds with an error?

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+
+    describe('quoteByPacket', function () {
```
emschwartz: What if the packet is invalid?

test/quoteSpec.js
```diff
@@ -0,0 +1,240 @@
...
+      this.response.data.data = { message: 'there was an error' }
+
+      const promise = ILQP._sendAndReceiveMessage(this.params)
+      this.plugin.emit('incoming_message', this.response)
+
+      yield expect(promise)
+        .to.eventually.be.rejectedWith(/there was an error/)
+    })
+
+    it('should time out without response', function * () {
```
emschwartz: This test should use fake timers

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+      this.validate(result)
+    })
+
+    it('should generate an id if one isn\'t provided', function () {
+      delete this.params.id
+      const result = Transport.createPacketAndCondition(this.params, 'psk')
+      this.validate(result)
+
+      const parsed = ILP.Packet.parse(result.packet)
+      assert.match(parsed.address, new RegExp(this.params.destinationAccount))
```
emschwartz: This test should specify exactly what the address should look like

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+      })
+    })
+
+    describe('PSK', function () {
+      it('should create packet and condition', function () {
+        // one field name is different
+        this.params.sharedSecret = this.params.secret
+
+        const result = ILP.PSK.createPacketAndCondition(this.params)
+        this.validate(result)
```
emschwartz: This should test against specific values

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+      this.params = {
+        id: 'ee39d171-cdd5-4268-9ec8-acc349666055',
+        secret: Buffer.from('shh_its_a_secret', 'base64')
+      }
+    })
+
+    it('should listen', function () {
+      const res = Transport.listen(this.plugin, this.params, () => {})
+      assert.isFunction(res, 'should return a function')
+    })
```
emschwartz: Need a test in case `plugin.connect` rejects or hangs

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+      yield Transport._validateTransfer(this.params)
+    })
+
+    it('should not accept transfer without condition', function * () {
+      delete this.params.transfer.executionCondition
+      yield expect(Transport._validateTransfer(this.params))
+        .to.be.rejectedWith(/no-execution/)
+    })
+
+    it('should not accept transfer for other account', function * () {
```
emschwartz: These tests should say "ignores" instead of "not accept" in cases where it's not rejecting the transfer

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+        'test.example.alice.~ipr.garbage'
+
+      yield expect(Transport._validateTransfer(this.params))
+        .to.be.rejectedWith(/not-my-packet/)
+    })
+
+    it('should not accept transfer for too little money', function * () {
+      this.params.transfer.amount = '0.1'
+      yield expect(Transport._validateTransfer(this.params))
+        .to.be.rejectedWith(/insufficient/)
```
emschwartz: This needs to test that the transfer is actually rejected

test/transportSpec.js
```diff
@@ -0,0 +1,279 @@
...
+      Transport.listen(this.plugin, this.params, this.callback, 'ipr')
+
+      // listener returns true for debug purposes
+      const res = yield this.plugin.emitAsync('incoming_prepare', this.transfer)
+      assert.isTrue(res[0])
+
+      yield this.fulfilled
+    })
+
+    it('should not fulfill when it generates the wrong fulfillment', function * () {
```
emschwartz: Instead of "not fulfill" it should reject the transfer

src/lib/ilqp.js
```diff
@@ -0,0 +1,180 @@
...
+      if (data.method === 'error') reject(data.data.message)
+
+      if (data.method === method) {
+        debug('response of type', method)
+        plugin.removeListener('incoming_message', onIncomingMessage)
+        resolve(response)
+      }
+    }
+
+    // TODO: optimize to not add a listener each time?
```
sharafian: If that doesn't cause it to stop working, I could just set the max listeners to something really high. It does at least clean up the listeners when the payments end.

src/lib/ilqp.js
```diff
@@ -0,0 +1,180 @@
...
+      account: connector,
+      data: {
+        method: 'quote_request',
+        data: quoteQuery
+      }
+    }
+  }).then((response) => {
+    return response.data.data
+  }).catch((err) => {
+    // TODO: should this do something else about errors?
```
sharafian: So I'll throw the response as an error?

src/lib/ilqp.js
```diff
@@ -0,0 +1,180 @@
...
+  destinationScale,
+  connectors,
+  timeout
+}) {
+  if (!xor(sourceAmount, destinationAmount)) {
+    throw new Error('should provide source or destination amount but not both' +
+      ' ' + JSON.stringify({ sourceAmount, destinationAmount }))
+  }
+
+  yield plugin.connect()
```
sharafian: The plugin does have a timeout option for connect, as part of the LPI. I could write a util function that does the promise.race external to the plugin, though, if you think that's better. One problem is that a ~10 second timeout isn't always enough.

src/lib/ilqp.js
emschwartz: I think plugins might want to set that themselves, but I don't think this module should call that method on the plugin

src/lib/ilqp.js
emschwartz: It's tricky because there are a couple of cases:

- multiple connectors, all return quotes - return the best quote
- multiple connectors, one throws an error while others return quotes - you want the quote and should ignore the errors
- multiple connectors, none respond - throw an error
- multiple connectors, multiple throw errors while none return quotes - ?

src/lib/packet.js
```diff
@@ -0,0 +1,64 @@
...
+function getFromTransfer (transfer) {
+  assert(transfer, 'transfer must be defined. got: ' + transfer)
+  assert(typeof transfer === 'object', 'got invalid transfer: ' + transfer)
+  assert(transfer.data, 'transfer missing data: ' + JSON.stringify(transfer))
+  assert(typeof transfer.data === 'object', 'got invalid transfer data: ' +
+    transfer.data)
+  assert(transfer.data.ilp_header, 'transfer.data missing ilp_header: ' +
+    JSON.stringify(transfer.data))
+
+  return transfer.data
```
sharafian: I want the packet to be the thing you attach to the `data` field (or the `ilp` field, in the future), so it'll be easier to switch to binary. We don't put anything else in `data` right now, so I think this should be OK.

src/lib/psk.js
```diff
@@ -0,0 +1,61 @@
...
+  const params = Object.assign({}, rawParams, { secret: rawParams.sharedSecret })
+  return Transport.createPacketAndCondition(params, 'psk')
+}
+
+/**
+  * Generate shared secret for Pre-Shared Key (PSK) transport protocol.
+  *
+  * @param {Object} params Parameters for creating PSK params
+  * @param {String} params.destinationAccount The ILP address that will receive PSK payments
+  * @param {String} params.secret secret used to generate the shared secret and the extra segments of destinationAccount
```
sharafian: Maybe something like `seed`?

src/lib/psk.js
```diff
@@ -0,0 +1,61 @@
...
+  * @param {String} params.id id id to distinguish from other receivers listening on the same PSK account. If supplied, it must be passed into the PSK.listen function's options.
+  *
+  * @return {PskParams}
+  */
+function generateParams ({
+  destinationAccount,
+  secret,
+  // TODO: best way to do receiverId?
```
sharafian: I'll remove it; it's not relevant anymore.

src/lib/spsp.js
```diff
@@ -68,77 +93,66 @@ const _quote = function * ({ plugin, spsp, sourceAmount, destinationAmount, ilp
...
+    id: payment.id,
+    // TODO: should connectorAccount become destinationAccount when
+    // there is no connector?
+    to: payment.connectorAccount || payment.destinationAccount,
+    amount: payment.sourceAmount,
+    data: packet,
+    executionCondition: condition,
+    expiresAt: moment()
+      .add(payment.sourceExpiryDuration, 'seconds')
+      .format()
```
sharafian: `format` formats to an ISO string

src/lib/transport.js
```diff
@@ -0,0 +1,190 @@
...
+  // TODO: best way to do receiverId?
+  secret,
+  allowOverPayment
+}, callback, protocol) {
+  assert(plugin && typeof plugin === 'object', 'plugin must be an object')
+  assert(typeof callback === 'function', 'callback must be a function')
+  assert(Buffer.isBuffer(secret), 'opts.secret must be a buffer')
+
+  // TODO: do this async
+  // yield plugin.connect()
```
sharafian: I'll change listen to be async as per your other comment, and then I'll be able to connect the plugin

src/lib/transport.js
```diff
@@ -0,0 +1,190 @@
...
+
+  const listener = co.wrap(autoFulfillCondition)
+  plugin.on('incoming_prepare', listener)
+
+  return function () {
+    plugin.removeListener('incoming_prepare', listener)
+  }
+}
+
+function * _validateTransfer ({
```
sharafian: `_validateOrRejectTransfer`?

test/quoteSpec.js
sharafian: https://github.com/interledgerjs/ilp/pull/79/files/2d71fcf1cc0b02fa9d167ecc295166476672bf55#diff-c50a802f4681c32c2a3e9ff6fa4360cfR194

test/quoteSpec.js
sharafian: I think that's the responsibility of the `Packet` tests

test/quoteSpec.js
sharafian: the timeout is configurable, so I can set it to something like 10 milliseconds. What's the reasoning for using fake timers?

src/lib/psk.js
emschwartz: `secretSeed`?

test/quoteSpec.js
emschwartz: Fair enough

src/lib/packet.js
emschwartz: This whole change should probably be coordinated with the update to the new packet format

src/lib/ilqp.js
emschwartz: What about having a default in the ILP client that makes it only wait for ~10 seconds and an option to disable that if you know you're connecting to a particularly slow ledger?

sharafian: maybe connection timeout ought to be a constructor property rather than one of the connect function, so that the user can set it without having to add a field to every function that calls connect

emschwartz: You mean plugin constructor option?

src/lib/ipr.js
```diff
@@ -19,6 +23,27 @@ function createPacketAndCondition (params) {
...
+  * @param {Function} params.fulfill async function that fulfills the transfer when it is called
+  */
+
+/**
+  * Listen on a plugin for incoming IPR payments, and auto-generate fulfillments.
+  *
+  * @param {Object} plugin Ledger plugin to listen on
+  * @param {Object} params Parameters for creating payment request
+  * @param {Buffer} params.secret Secret to generate fulfillments with
+  * @param {Buffer} [params.allowOverPayment] Accept payments with higher amounts than expected
```
emschwartz: If it's optional you should also specify the default value

src/lib/transport.js
```diff
@@ -46,20 +51,18 @@ function createPacketAndCondition ({
...
-  // TODO: best way to do receiverId?
+function * listen (plugin, {
   secret,
   allowOverPayment
 }, callback, protocol) {
   assert(plugin && typeof plugin === 'object', 'plugin must be an object')
   assert(typeof callback === 'function', 'callback must be a function')
   assert(Buffer.isBuffer(secret), 'opts.secret must be a buffer')
 
-  // TODO: do this async
```
emschwartz: I think this should use a timeout by default

test/quoteSpec.js
```diff
@@ -72,6 +72,14 @@ describe('ILQP', function () {
         this.result)
     })
 
+    it('should throw if getInfo returns no connectors', function * () {
```
emschwartz: "be rejected" not "throw"

test/transportSpec.js
```diff
@@ -148,69 +158,99 @@ describe('Transport', function () {
...
+    it('should ignore transfer without condition', function * () {
       delete this.params.transfer.executionCondition
-      yield expect(Transport._validateTransfer(this.params))
-        .to.be.rejectedWith(/no-execution/)
+      assert.equal(
+        yield Transport._validateOrRejectTransfer(this.params),
+        'no-execution')
     })
 
     it('should not accept transfer for other account', function * () {
```
emschwartz: "should ignore"

test/transportSpec.js
```diff
@@ -148,69 +158,99 @@ describe('Transport', function () {
...
+        yield Transport._validateOrRejectTransfer(this.params),
+        'overpayment')
+      yield this.rejected
     })
 
     it('should accept extra money with "allowOverPayment"', function * () {
       this.params.transfer.amount = '1.1'
       this.params.allowOverPayment = true
-      yield Transport._validateTransfer(this.params)
+      // no error-code is returned on success
```
emschwartz: You could just make it return true for testing purposes

test/transportSpec.js
```diff
@@ -248,13 +295,13 @@ describe('Transport', function () {
       yield this.fulfilled
     })
 
-    it('should not fulfill when it generates the wrong fulfillment', function * () {
+    it('should reject when it generates the wrong fulfillment', function * () {
```
emschwartz: It would also be good to include a test to make sure that when the packet is modified slightly it will fail. It's kind of similar to this one but in this case the execution condition isn't even valid

test/transportSpec.js
sharafian: autoFulfillCondition checks truthyness to make sure an error code isn't being returned

src/lib/transport.js
```diff
@@ -0,0 +1,223 @@
...
+  assert(typeof destinationAmount === 'string', 'destinationAmount must be a string')
+  assert(typeof destinationAccount === 'string', 'destinationAccount must be a string')
+  assert(Buffer.isBuffer(secret), 'secret must be a buffer')
+
+  const receiverId = base64url(cryptoHelper.getReceiverId(secret))
+  const address = destinationAccount + '.~' + protocol + '.' + receiverId +
+    (id ? ('.' + id) : '')
+
+  const blobData = { expiresAt, data }
+  const blob = data &&
```
michielbdejong: this is a bug; if data is undefined, that doesn't mean blob has to be. at least that's what you just said yourself ;)

src/lib/ilqp.js
```diff
@@ -0,0 +1,202 @@
...
+  const quoteQuery = omitUndefined({
+    source_address: plugin.getAccount(),
+    source_amount: sourceAmount,
+    destination_address: destinationAddress,
+    destination_amount: destinationAmount,
+    destination_expiry_duration: destinationExpiryDuration,
+  })
+
+  const quoteConnectors = connectors || plugin.getInfo().connectors || []
+  debug('quoting', amount, 'via', quoteConnectors)
```
emschwartz: Should log the destination account and whether that amount is the source or destination amount

src/lib/spsp.js
```diff
@@ -68,77 +93,65 @@ const _quote = function * ({ plugin, spsp, sourceAmount, destinationAmount, ilp
...
-      },
-      executionCondition: request.condition,
-      expiresAt: request.expires_at
-    })
-
-    return { fulfillment }
+  yield plugin.sendTransfer({
+    id: payment.id,
+    // TODO: should connectorAccount become destinationAccount when
+    // there is no connector?
```
emschwartz: If the only reason that would happen is if it's a same-ledger transfer, then yes

src/lib/transport.js
```diff
@@ -0,0 +1,260 @@
...
+const base64url = require('../utils/base64url')
+const BigNumber = require('bignumber.js')
+const { safeConnect, omitUndefined } = require('../utils')
+
+function _safeDecrypt (data, secret) {
+  if (!data) return {}
+  try {
+    return cryptoHelper.aesDecryptObject(data, secret)
+  } catch (err) {
+    debug('decryption error:', err.message)
```
emschwartz: Maybe log the data as well for debugging purposes

src/lib/transport.js
```diff
@@ -0,0 +1,260 @@
...
+      ' transfer=' + transfer.amount +
+      ' packet=' + destinationAmount)
+    return yield _reject(plugin, transfer.id, {
+      code: 'S03',
+      name: 'Invalid Amount',
+      message: 'got notification of transfer where amount is more than expected'
+    })
+  }
+
+  if (expiresAt && moment().isAfter(expiresAt)) {
```
emschwartz: Since we're not really using much of `moment`'s advanced functionality I think it might be worth removing that dependency (doesn't have to be done in this PR though)

src/utils/packet.js
```diff
@@ -0,0 +1,37 @@
...
+  } catch (e) {
+    debug('error while parsing packet: ' + e.message)
+    return undefined
+  }
+}
+  
+function getFromTransfer (transfer) {
+  assert(transfer, 'transfer must be defined. got: ' + transfer)
+  assert(typeof transfer === 'object', 'got invalid transfer: ' + transfer)
+  assert(Buffer.isBuffer(transfer.ilp || transfer.data),
```
emschwartz: I think `transfer.ilp` is a string

src/utils/packet.js
```diff
@@ -0,0 +1,37 @@
...
+    'transfer.ilp or transfer.data must be buffer')
+
+  return transfer.ilp || transfer.data
```
emschwartz: I don't see the point in possibly returning `transfer.data`. We're not actually making this change backwards compatible and the updated spec says the ILP packet should be under the `ilp` key.

Having this small wrapper around the `ilp-packet` module doesn't make that much sense to me. Why not just use the `ilp-packet` class directly where these functions are being used?

test/quoteSpec.js
emschwartz: Still not sure how this is testing what it says

test/quoteSpec.js
```diff
@@ -0,0 +1,247 @@
...
+        connectors: [ 'test.example.connie' ]
+      }
+      this.result = {
+        sourceAmount: '1',
+        destinationAmount: '1',
+        connectorAccount: 'test.example.connie',
+        sourceExpiryDuration: '0'
+      }
+
+      this.plugin.sendMessage = (msg) => {
```
emschwartz: The problem I see with using this custom setup instead of something like sinon is that you're not actually checking what the arguments are that are passed to this mock

test/quoteSpec.js
```diff
@@ -0,0 +1,247 @@
...
+    it('should return a local quote if destination is local', function * () {
+      this.params.destinationAddress = 'test.example.bob'
+      const response = yield ILQP.quote(this.plugin, this.params)
+
+      delete this.result.connectorAccount // local quote has no connector
+      assert.deepEqual(response,
+        this.result)
+    })
+
+    it('should throw if source and dest amounts are defined', function * () {
```
emschwartz: The tests should say "be rejected" instead of "throw" if that's what they do. Looking at the test output would give you a mistaken idea of what the functionality is actually expected to be (this is true of a bunch of these tests)

test/quoteSpec.js
emschwartz: Wouldn't running this on a super fast machine make this test fail right now?

test/spspSpec.js
```diff
@@ -96,170 +93,78 @@ describe('SPSP Module', function () {
...
-      delete this.payment.destinationAmount
-      yield expect(SPSP.sendPayment(this.plugin, this.payment))
-        .to.eventually.be.rejectedWith(/missing destinationAmount/)
-    })
-
-    it('should fail without sourceAmount', function * () {
-      delete this.payment.sourceAmount
-      yield expect(SPSP.sendPayment(this.plugin, this.payment))
-        .to.eventually.be.rejectedWith(/missing sourceAmount/)
-    })
```
emschwartz: Why are all these tests being taken out?

test/transportSpec.js
```diff
@@ -0,0 +1,356 @@
...
+    it('should reject when packet details have been changed', function * () {
+      this.transfer.ilp = Buffer.concat([this.transfer.ilp, Buffer.from('00', 'hex')])
+      yield Transport.listen(this.plugin, this.params, this.callback, 'ipr')
+
+      // listener returns false for debug purposes
+      yield this.plugin.emitAsync('incoming_prepare', this.transfer)
+      yield this.rejected
+    })
+
+    it('should pass the fulfill function to the callback', function * () {
```
emschwartz: Nitpicky but could you say all the things it's actually supposed to pass to the callback?

test/transportSpec.js
```diff
@@ -0,0 +1,356 @@
...
+        assert.isString(details.destinationAccount, 'must pass in account')
+        assert.isString(details.destinationAmount, 'must pass in amount')
+        assert.isFunction(details.fulfill, 'fulfill callback must be a function')
+        details.fulfill()
+      }
+
+      yield Transport.listen(this.plugin, this.params, this.callback, 'ipr')
+      yield this.plugin.emitAsync('incoming_prepare', this.transfer)
+      yield fulfilled
+    })
```
emschwartz: Could you add a test for what happens if the callback throws?

src/lib/ilqp.js
```diff
@@ -0,0 +1,202 @@
...
+  plugin,
+  method,
+  message,
+  timeout
+}) {
+  const id = message.data.id = message.data.id || uuid()
+  debug('sending message:', JSON.stringify(message))
+
+  const responded = new Promise((resolve, reject) => {
+    function onIncomingMessage (response) {
```
sharafian: the function definition is so that it can refer to itself when removing the `incoming_message` listener

src/lib/transport.js
sharafian: Fair enough. The methods we need could be written in utils.

src/utils/packet.js
sharafian: the `ilp-packet` module returns a buffer, and parses buffers. Shouldn't the LPI use them too?

src/utils/packet.js
sharafian: Removing the support for `transfer.data`. I guess having a wrapper around the packet functionality isn't strictly required, but it's been extremely useful throughout this rewrite (switching from JSON to binary, switching from LPI versions, etc.) so I'm inclined to keep it around for the future.

test/quoteSpec.js
sharafian: Added some assert statements in there to make sure the data is OK.

test/quoteSpec.js
sharafian: It never emits an `incoming_message`, so it times out no matter what.

test/quoteSpec.js
sharafian: I delete the connectors getting passed it manually, so it has to get them from the getInfo function. I have another test that tells getInfo to not return anything and then also doesn't pass in connectors, to make sure that fails.

test/spspSpec.js
sharafian: https://github.com/interledgerjs/ilp/pull/79/files#diff-b557f2459a4c96ddfe2567537174fa7aR100 it's just checking all these asserts to make sure the arguments are there.

src/utils/packet.js
```diff
@@ -0,0 +1,35 @@
...
+  } catch (e) {
+    debug('error while parsing packet: ' + e.message)
+    return undefined
+  }
+}
+
+function getFromTransfer (transfer) {
+  assert(transfer, 'transfer must be defined. got: ' + transfer)
+  assert(typeof transfer === 'object', 'got invalid transfer: ' + transfer)
+  assert(Buffer.isBuffer(transfer.ilp || transfer.data),
```
sentientwaffle: Should `transfer.data` be removed?

src/lib/transport.js
```diff
@@ -0,0 +1,274 @@
...
+        ' executionCondition=' + transfer.executionCondition +
+        ' our condition=' + cc.toCondition(preimage))
+      return yield _reject(plugin, transfer.id, {
+        code: 'S05',
+        name: 'Wrong Condition',
+        message: 'receiver generated a different condition from the transfer'
+      })
+    }
+
+    const parsed = Packet.parseFromTransfer(transfer)
```
sentientwaffle: `parsed` might be undefined

src/lib/spsp.js
```diff
@@ -36,19 +39,41 @@ const _querySPSP = function * (receiver) {
...
-const _quote = function * ({ plugin, spsp, sourceAmount, destinationAmount, ilp }) {
-  if (!plugin) throw new Error('missing plugin')
-  if (!spsp.destination_account) throw new Error('missing destination account')
-  if (!spsp.maximum_destination_amount) throw new Error('missing maximum destination amount')
-  if (!spsp.minimum_destination_amount) throw new Error('missing minimum destination amount')
+const query = co.wrap(_querySPSP)
 
-  const client = new IlpCore.Client(plugin, ilp)
-  const quote = yield client.quote({
+const quote = function * (plugin, {
```
sentientwaffle: Are the `spsp` functions going to expose the integer amounts? The implementation in https://github.com/interledgerjs/ilp/pull/80 doesn't (see https://github.com/interledgerjs/ilp/pull/80/files#diff-b557f2459a4c96ddfe2567537174fa7aR49)

sharafian: I think the actual SPSP server and the ILQP requests should be integer amounts. But the SPSP functions that the user calls could use decimal amounts.

src/utils/packet.js
```diff
@@ -0,0 +1,35 @@
...
+  } catch (e) {
+    debug('error while parsing packet: ' + e.message)
+    return undefined
+  }
+}
+
+function getFromTransfer (transfer) {
+  assert(transfer, 'transfer must be defined. got: ' + transfer)
+  assert(typeof transfer === 'object', 'got invalid transfer: ' + transfer)
+  assert(Buffer.isBuffer(transfer.ilp),
```
emschwartz: According to https://github.com/interledger/rfcs/pull/165 `transfer.ilp` will be a string

test/transportSpec.js
```diff
@@ -0,0 +1,382 @@
...
+
+    it('should reject if the listen callback rejects', function * () {
```
emschwartz: This test description makes me think that if the callback returns a Promise that rejects that would also work. Is that true?

sharafian: yeah

test/quoteSpec.js
```diff
@@ -0,0 +1,253 @@
...
+
+    it('should reject if getInfo returns no connectors', function * () {
+      delete this.params.connectors
+      this.plugin.getInfo = () => ({ prefix: 'test.example.' })
+
+      yield expect(ILQP.quote(this.plugin, this.params))
+        .to.be.rejectedWith(/no connectors specified/)    
+    })
+
+    it('should quote by destination amount', function * () {
```
emschwartz: What about quoting by source amount?

sharafian: that's what the rest of them are

src/lib/spsp.js
```diff
@@ -68,77 +110,71 @@ const _quote = function * ({ plugin, spsp, sourceAmount, destinationAmount, ilp
...
-    const quote = yield _quote({ plugin, spsp, destinationAmount: amount, ilp })
-    return _createPayment(spsp, quote)
+  const fulfill = new Promise((resolve, reject) => {
+    function remove () {
+      plugin.removeListener('outgoing_fulfill', fulfill)
+      plugin.removeListener('outgoing_cancel', cancel)
+      plugin.removeListener('outgoing_reject', cancel)
+    }
+
+    function fulfill (transfer, fulfillment) {
```
sharafian: renamed outer promise to `listen`

src/lib/transport.js
```diff
@@ -37,17 +40,24 @@ function createPacketAndCondition ({
   const address = destinationAccount + '.~' + protocol + '.' + receiverId +
     (id ? ('.' + id) : '')
```
emschwartz: Since we're already using the payment `id` to ensure that the hashes are different every time, it seems like we might as well use that as a nonce in the generation of the encryption key as well. 

I think we should move the id out of the ILP address and into the transport `unsafeHeaders` section. That ID isn't used for routing and isn't useful for anyone aside from the sender and receiver.

@justmoon thoughts?

docs/README.template.md
```diff
@@ -1,11 +1,11 @@
 <h1 align="center">
   <a href="https://interledger.org"><img src="ilp_logo.png" width="150"></a>
   <br>
-  ILP Client
+  ILP
 </h1>
 
 <h4 align="center">
-A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver library
+A low-level JS <a href="https://interledger.org">Interledger</a> library
```
emschwartz: It might be more accurate to say "The Javascript client library for Interledger"

docs/README.template.md
```diff
@@ -23,26 +23,50 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
 [snyk-image]: https://snyk.io/test/npm/ilp/badge.svg
 [snyk-url]: https://snyk.io/test/npm/ilp
 
-This is a low-level interface to ILP, largely intended for building ILP into other [Application layer](https://github.com/interledger/rfcs/tree/master/0001-interledger-architecture) protocols.
+This module bundles low-level and high-level interfaces to ILP, largely intended for building ILP into other [Application layer](https://github.com/interledger/rfcs/tree/master/0001-interledger-architecture) protocols.
```
emschwartz: That's confusing. I would just take this line out now

README.md
```diff
@@ -23,26 +23,50 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
...
-* Amount negotiation
-* Communication of requests from recipient to sender
-
 ## Installation
 
 `npm install --save ilp ilp-plugin-bells`
 
 *Note that [ledger plugins](https://www.npmjs.com/search?q=ilp-plugin) must be installed alongside this module*
 
+## Simple Payment Setup Protocol (SPSP)
```
emschwartz: This should link to the spec

src/utils/crypto.js
```diff
@@ -39,24 +39,34 @@ function hmacJsonForPskCondition (obj, sharedSecret) {
...
+}
+
+// turn buffer into encrypted buffer
+function aesEncryptBuffer (buffer, sharedSecret) {
   const pskEncryptionKey = hmac(sharedSecret, PSK_ENCRYPTION_STRING)
   const cipher = crypto.createCipher('aes-256-ctr', pskEncryptionKey)
 
   return Buffer.concat([
-    cipher.update(Buffer.from(JSON.stringify(obj), 'utf8')),
+    cipher.update(buffer),
```
sharafian: https://nodejs.org/api/crypto.html#crypto_class_cipher the documentation on the cipher class says the complete ciphertext is `cipher.update(buffer) + cipher.final()`

src/utils/crypto.js
```diff
@@ -39,24 +39,34 @@ function hmacJsonForPskCondition (obj, sharedSecret) {
...
-function aesDecryptObject (encrypted, sharedSecret) {
+// turn buffer into decrypted buffer
+function aesDecryptBuffer (encrypted, sharedSecret) {
   const pskEncryptionKey = hmac(sharedSecret, PSK_ENCRYPTION_STRING)
   const decipher = crypto.createDecipher('aes-256-ctr', pskEncryptionKey)
 
-  const decoded = Buffer.concat([
-    decipher.update(Buffer.from(encrypted, 'base64')),
+  return Buffer.concat([
+    decipher.update(encrypted),
```
sharafian: https://nodejs.org/api/crypto.html#crypto_class_decipher that's not how the decipher class works, according to the docs

test/utilSpec.js
```diff
@@ -12,4 +13,31 @@ describe('Utils', function () {
         .throws(() => Crypto.aesDecryptObject('garbage', Buffer.from('trash', 'base64')))
     })
   })
+
+  describe('details', function ()  {
```
sharafian: The transport tests already use the class fully; the utils tests are just to catch the edge cases not covered by the other tests

src/lib/transport.js
```diff
@@ -27,6 +28,8 @@ function createPacketAndCondition ({
   destinationAccount,
   secret,
   data,
+  headers,
+  unsafeHeaders,
```
sharafian: The parameters of createPacketAndCondition in IPR and PSK are passed directly through to the Transport class. I'll add a test to make sure that additional headers can be specified though.

README.md
sharafian: speaking of which, we need to put the new SPSP in the rfcs

src/lib/transport.js
emschwartz: The header parameters should also be included in the JSDoc of the IPR and PSK functions

src/utils/details.js
```diff
@@ -0,0 +1,100 @@
...
+  const headerLines = Object.keys(headers)
+    .map((k) => k + ': ' + headers[k])
+    .join('\n') + DATA_DELIMITER
+
+  let rawData = data
+  if (!Buffer.isBuffer(data) && typeof data === 'object') {
+    rawData = Buffer.from(JSON.stringify(data), 'utf8')
+  } else if (data === undefined) {
+    rawData = Buffer.from([])
+  }
```
sharafian: I think those should be set manually by the application layer protocol, because the transport layer doesn't care about the data inside the private PSK request.

src/utils/details.js
```diff
@@ -1,7 +1,10 @@
 'use strict'
 
 const cryptoHelper = require('./crypto')
+const Packet = require('./packet')
 const DATA_DELIMITER = '\n\n'
+const STATUS_LINE_REGEX = /^PSK\/1\.0 (PUBLIC|PRIVATE)$/
```
emschwartz: Should it say `PSK 1.0` even when it's being used for IPR?

docs/README.template.md
```diff
@@ -55,47 +77,59 @@ This library handles the generation of payment requests, but **not the communica
...
+    console.log(transfer.id, 'was fulfilled with', fulfillment)
+    stopListening()
+  })
 }).catch((err) => {
   console.log(err)
 })
-
 ```
 
 ### Pre-Shared Key (PSK) Transport Protocol
```
emschwartz: This and the IPR headers should also link to their respective specs

src/utils/details.js
sharafian: I think it should, but that might be more of a point for the discussion of whether IPR and PSK should be different at all. I think right now I'm leaning towards them being two separate protocols in function names and in flows only, but having all data (ilp addresses, packet data, etc.) be marked PSK.

docs/README.template.md
sharafian: linked IPR, will link PSK once my PR to RFCs is merged

docs/README.template.md
```diff
@@ -23,30 +23,52 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
...
+co(function * () {
+  const payment = yield SPSP.quote(plugin, {
+    receiver: 'bob@blue.ilpdemo.org'
+    sourceAmount: '1',
+  })
+
+  console.log('got SPSP payment details:', payment)
 
-## Interledger Payment Request (IPR) Transport Protocol
+  const { fulfillment } = yield SPSP.sendPayment(plugin, payment)
```
justmoon: As a developer I'm reading this thinking "What's fulfillment? What do I do with it?"

The answer is nothing. So maybe lets not use the return value at all in the example?

If you take out the `fulfillment =` bit, I would add a comment or change the console.log to make it clear that the payment is guaranteed to have been received by the recipient by this point.

docs/README.template.md
```diff
@@ -55,47 +77,59 @@ This library handles the generation of payment requests, but **not the communica
...
 })
 
 co(function * () {
-  yield receiver.listen()
-  receiver.on('incoming', (transfer, fulfillment) => {
-    console.log('received transfer:', transfer)
-    console.log('fulfilled transfer hold with fulfillment:', fulfillment)
+  const stopListening = yield ILP.IPR.listen(receiver, {
+    secret: Buffer.from('secret', 'utf8')
+  }, (params) => {
```
justmoon: Maybe `({ fulfill }) => {`

docs/README.template.md
```diff
@@ -23,30 +23,52 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
...
+const co = require('co')
+const SPSP = require('ilp').SPSP
+const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
+
+const plugin = new FiveBellsLedgerPlugin({
+  account: 'https://red.ilpdemo.org/ledger/accounts/alice',
+  password: 'alice'
+})
+
+co(function * () {
```
justmoon: Since async/await has landed in Node.js I think it would be fine to use it in the example.

I do realize that not all JavaScript developers are familiar with it yet, but it's going to spread quickly and we want this to look modern. Maybe include a link to a page that explains async/await somewhere.

justmoon: Could also have examples for both ES6 with `import` and `async/await` and ES5/CommonJS with `require` and `co` or maybe even just plain promises.

There is a way to make collapsible sections in Github Markdown: https://github.com/dear-github/dear-github/issues/166#issuecomment-252282315

So we could have the ES6 code example and then an ES5 example in a collapsible block below.

src/lib/transport.js
justmoon: Agreed. That's what I did as well. (When creating the example packet for the ilp-packet-viewer.)

src/lib/transport.js
```diff
@@ -37,17 +40,24 @@ function createPacketAndCondition ({
   const address = destinationAccount + '.~' + protocol + '.' + receiverId +
     (id ? ('.' + id) : '')
 
-  const blobData = omitUndefined({
-    expires_at: expiresAt,
-    data: data
-  })
+  const details = createDetails({
+    unsafeHeaders: Object.assign({
+      'Key-Id': receiverId, // TODO: is this right?
```
justmoon: I think I prefer one of these:

``` http
Key: hmac-sha-256 W2WuqXqCedW3hj4UsQdzBg
```

Here are some examples from HTTP:

``` http
Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==
Content-Type: text/html; charset=utf-8
Content-Disposition: attachment; filename="fname.ext"
Cookie: $Version=1; Skin=new;
Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43
```

Authorization seems like the closest analog and it's a header a lot of developers have dealt with.

[Non-standard](http://progrium.com/blog/2012/12/17/http-signatures-with-content-hmac/), but when I saw it I thought it was very elegant and simple:

``` http
Content-HMAC: <hash mechanism> <base64 encoded binary HMAC>
```

Example:
```
Content-HMAC: sha1 f1wOnLLwcTexwCSRCNXEAKPDm+U=
```

Saves 18 bytes, still easy to parse, still easy enough to guess that hmac-sha-256 is some kind of type and the base64 thingy is the actual key. Can't really come up with a good reason not to use this kind of format.

sharafian: From the article you linked it looks like `Content-HMAC` contains the digest of the data, rather than information on an HMAC key.

Also, did we decide that the second part of `hmac-sha-256 W2WuqXqCedW3hj4UsQdzBg` is going to be used as a nonce? It sounds like we're keeping the receiverId in the ILP address.

emschwartz: `receiverId` should stay in the ILP address and the UUID that we were previously appending to the address should be moved to this header.

It makes sense for that header to be something like `Key: hmac-sha-256 a41d1a2f-632d-47b4-b730-3070270190ea` (or _maybe_ `Key-Gen`, since it's about how you generate the payment key). As @justmoon points out there's precedent for including two fields in the header value.

I think the thing after `hmac-sha-256` should be a nonce that the sender and recipient will use to generate the (unique) payment key (`HMAC(shared secret, nonce)`), which will be used for encryption and generating the fulfillment (`HMAC(payment key, ILP packet)`). We were already using it to ensure that the conditions were unique, so (I think) we might as well use a unique key for encryption as well.

sharafian: Another reason to use the uuid as a nonce is to ensure that the same private headers/data won't encrypt to the same cipher-text across two different payments.

README.md
```diff
@@ -92,14 +90,14 @@ const receiver = new FiveBellsLedgerPlugin({
...
+  const stopListening = await ILP.IPR.listen(receiver, {
     secret: Buffer.from('secret', 'utf8')
-  }, (params) => {
-    console.log('got transfer:', params.transfer)
+  }, ({ transfer, fulfill }) => {
+    console.log('got transfer:', transfer)
 
     console.log('fulfilling.')
-    return params.fulfill()
+    return fulfill()
```
emschwartz: I think it might be worth making this an async function so that you can `await fulfill()` and then log a statement afterwards that indicates the money has been received. It would also be good to replace the log line above with something more descriptive of what the recipient is doing

src/utils/details.js
```diff
@@ -36,45 +38,49 @@ function createDetails ({
...
+
   const privateRequest = _createRequest({
-    method: 'PRIVATE',
+    statusLine: false,
     headers,
     data
   })
 
-  const encrypted = cryptoHelper.aesEncryptBuffer(privateRequest, secret)
+  const encrypted = cryptoHelper.aesEncryptBuffer(privateRequest, paymentKey)
```
emschwartz: Nitpick: I think it's more common to put the key first

src/utils/details.js
```diff
@@ -36,45 +38,49 @@ function createDetails ({
...
   const publicRequest = _createRequest({
-    method: 'PUBLIC',
-    headers: unsafeHeaders,
+    statusLine: true,
+    headers: Object.assign({
+      'Key': 'hmac-sha-256 ' + base64url(token)
+    }, unsafeHeaders),
```
emschwartz: If you set a `Key` header you would make the packet unreadable because this won't actually respect the Key you set for encrypting the packet and the recipient won't know how to generate the decryption key. Either it shouldn't allow you to set the `Key` header or it should use the token you provide to generate the payment key

src/utils/details.js
```diff
@@ -84,9 +90,26 @@ function parseDetails ({
...
   const detailsBuffer = Buffer.from(details, 'base64')
-  const publicRequest = _parseRequest(detailsBuffer)
-  const decrypted = cryptoHelper.aesDecryptBuffer(publicRequest.data, secret)
-  const privateRequest = _parseRequest(decrypted)
+  const publicRequest = _parseRequest({
+    request: detailsBuffer,
+    statusLine: true
+  })
+
+  const [ , token ] = publicRequest.headers['key'].match(KEY_HEADER_REGEX)
```
emschwartz: I think if this doesn't match, this line will throw `TypeError: Cannot read property 'Symbol(Symbol.iterator)' of null`

src/utils/details.js
```diff
@@ -84,9 +90,26 @@ function parseDetails ({
...
+  if (!token) {
+    throw new Error('invalid Key header in',
+      JSON.stringify(publicRequest.headers))
+  }
+
+  const paymentKey = cryptoHelper.getPaymentKey(
+    secret,
+    Buffer.from(token, 'base64'))
+
+  const decrypted = cryptoHelper.aesDecryptBuffer(publicRequest.data, paymentKey)
```
emschwartz: Can you check to make sure the way this is done will work even with the largest possible data field?

src/utils/crypto.js
```diff
@@ -4,44 +4,28 @@ const crypto = require('crypto')
 const stringify = require('canonical-json')
```
emschwartz: This dependency should be removed

src/lib/transport.js
```diff
@@ -0,0 +1,290 @@
...
+  })
+
+  const packet = Packet.serialize({
+    account: address,
+    amount: destinationAmount,
+    data: details
+  })
+
+  const condition = base64url(cc.toCondition(
+    cryptoHelper.hmacJsonForPskCondition(
```
emschwartz: This isn't right. Why is it using the hmacJson function? This should hmac the packet buffer. The hmacJson method should not be used anymore and should be removed (along with the `canonical-json` dependency)

This is also why it's important to have a test to match the output against a specific output. The current tests would not catch this because the sender and recipient are consistent with one another but they don't match the spec.

src/utils/crypto.js
```diff
@@ -57,37 +41,34 @@ function hmacJsonForPskCondition (obj, sharedSecret) {
   return hmacDigest
 }
 
-// turn object into encrypted buffer
-function aesEncryptObject (obj, sharedSecret) {
+// turn buffer into encrypted buffer
+function aesEncryptBuffer (sharedSecret, buffer) {
   const pskEncryptionKey = hmac(sharedSecret, PSK_ENCRYPTION_STRING)
   const cipher = crypto.createCipher('aes-256-ctr', pskEncryptionKey)
```
emschwartz: Can you put the `aes-256-ctr` at the top of the file as a constant?

package.json
```diff
@@ -38,11 +38,12 @@
   "dependencies": {
     "aguid": "^1.0.4",
     "bignumber.js": "^2.4.0",
-    "canonical-json": "0.0.4",
     "debug": "^2.2.0",
     "eventemitter2": "^2.0.0",
     "ilp-core": "^11.0.0",
```
emschwartz: `ilp-core` and `aguid` are no longer a dependencies and `eventemitter2` is only a `devDependency`

src/lib/transport.js
```diff
@@ -0,0 +1,290 @@
...
+      protocol,
+      allowOverPayment,
+      secret
+    })
+
+    if (err) return err
+
+    const preimage = cryptoHelper.hmacPacketForPskCondition(
+      Packet.getFromTransfer(transfer),
+      secret)
```
emschwartz: Shouldn't this be using the payment key?

sharafian: It doesn't seem like there's a difference in security either way. What's the reasoning for using the payment key?

src/lib/transport.js
```diff
@@ -0,0 +1,290 @@
...
+  destinationAccount,
+  secret,
+  data,
+  headers,
+  unsafeHeaders,
+  expiresAt
+}, protocol) {
+  assert(typeof destinationAmount === 'string', 'destinationAmount must be a string')
+  assert(typeof destinationAccount === 'string', 'destinationAccount must be a string')
+  assert(Buffer.isBuffer(secret), 'secret must be a buffer')
```
sharafian: I don't think so. This is a low level library, so people using it will invariably have opportunities to do insecure things if they want to. It's more likely to get in the way of people testing or developing things on top of this.

src/lib/psk.js
```diff
@@ -0,0 +1,94 @@
...
+  secretSeed
+}) {
+  assert(typeof destinationAccount === 'string', 'destinationAccount must be a string')
+  assert(Buffer.isBuffer(secretSeed), 'secretSeed must be a buffer')
+
+  const token = base64url(cryptoHelper.getPskToken(secretSeed))
+  const sharedSecret =
+    base64url(cryptoHelper.getPskSharedSecret(secretSeed, token))
+
+  const receiverId = base64url(cryptoHelper.getReceiverId(sharedSecret))
```
emschwartz: I think lines 49 - 53 should be moved into the crypto helper, which could just export a function `generatePskParams(secretSeed) => {sharedSecret, receiverId, token}` or something like that. I think that would make some of the crypto easier to follow

src/utils/crypto.js
```diff
@@ -1,47 +1,31 @@
...
-  function hmacJsonForIprCondition (obj) {
-    const jsonString = stringify(obj)
-    return hmac(iprConditionKey, jsonString)
-  }
-
-  function getReceiverId () {
-    return hmac(hmacKey, IPR_RECEIVER_ID_STRING).slice(0, 8)
-  }
+function getPskToken () {
+  return crypto.randomBytes(16)
```
emschwartz: Let's make this length a named constant at the top of the file too

src/utils/condition.js
```diff
@@ -3,21 +3,18 @@
 const crypto = require('crypto')
 const base64url = require('./base64url')
 
-function toConditionUri (conditionPreimage) {
+function toCondition (conditionPreimage) {
```
emschwartz: Can we move these functions into the crypto helper?

Also, I think it should export a `generateCondition(sharedSecret, packet)` function instead of `toCondition` from the fulfillment.

src/utils/crypto.js
```diff
@@ -50,44 +34,40 @@ function hmac (key, message) {
   return h.digest()
 }
 
-function hmacJsonForPskCondition (obj, sharedSecret) {
+function hmacPacketForPskCondition (packet, sharedSecret) {
```
emschwartz: To make this even more clear what it's for I think we should rename this to `generateFulfillment`.

README.md
```diff
@@ -272,218 +276,5 @@ Parameters for an SPSP payment
...
-### Receiver~PskParams : <code>Object</code>
-**Kind**: inner typedef of <code>[Receiver](#module_Receiver)</code>  
-**Properties**
-
-| Name | Type | Description |
-| --- | --- | --- |
-| destinationAccount | <code>string</code> | Receiver's ILP address |
-| sharedSecret | <code>string</code> | Base64Url-encoded shared secret |
+ERROR, Cannot find module.
+ERROR, Cannot find module.
```
emschwartz: Womp womp...

test/transportSpec.js
```diff
@@ -0,0 +1,518 @@
...
+        // the data is still encrypted, so we can't check it from just parsing
+        assert.isString(data)
+        assert.equal(amount, '1')
+        assert.match(account, /^test\.example\.alice\./)
+      }
+    })
+
+    it('should create a valid packet and condition', function () {
+      const result = Transport.createPacketAndCondition(this.params, 'psk')
+      this.validate(result)
```
emschwartz: This needs to test exactly what the result looks like. Just checking that the account matches `/^test\.example\.alice\./` is insufficient because it is supposed to add to the account in a way that is important for the protocol to work properly

test/transportSpec.js
```diff
@@ -0,0 +1,518 @@
...
+      this.params = {
+        destinationAccount: 'test.example.alice',
+        secretSeed: Buffer.from('shh_its_a_secret', 'base64')
+      }
+    })
+
+    it('should generate psk params', function () {
+      const params = ILP.PSK.generateParams(this.params)
+
+      assert.match(params.destinationAccount, /^test\.example\.alice/)
```
emschwartz: This needs to check the full address, not just the prefix that is supplied in the params

emschwartz: This should also check that the way it generates the shared secret is consistent with the spec. It should test against a hard-coded value, possibly using reRequire to mock crypto.randomBytes so that it always returns the same thing in testing.

src/lib/spsp.js
```diff
@@ -36,19 +43,76 @@ const _querySPSP = function * (receiver) {
     .set('Accept', 'application/json')).body
 }
```
emschwartz: This should probably validate the response and let you know if it's a valid SPSP receiver

src/lib/spsp.js
```diff
@@ -36,19 +43,76 @@ const _querySPSP = function * (receiver) {
...
+  * @param {Object} plugin Ledger plugin used for quoting.
+  * @param {Object} params Quote parameters
+  * @param {String} params.receiver webfinger account identifier (eg. 'alice@example.com') or URL to SPSP endpoint.
+  * @param {String} [params.sourceAmount] source amount to quote. This is a decimal, NOT an integer. It will be shifted by the sending ledger's scale to get the integer amount.
+  * @param {String} [params.destinationAmount] destination amount to quote. This is a decimal, NOT an integer. It will be shifted by the receiving ledger's scale to get the integer amount.
+  * @param {Array} [params.connectors=[]] connectors to quote. These will be supplied by plugin.getInfo if left unspecified.
+  * @param {String} [params.id=uuid()] id to use for payment. sending a payment with the same id twice will be idempotent. If left unspecified, the id will be generated randomly.
+  * @param {Number} [params.timeout=5000] how long to wait for a quote response (ms).
+  *
+  * @returns {Promise<SpspPayment>} SPSP payment object to be sent.
```
emschwartz: The `SpspPayment` type should be documented

src/lib/spsp.js
```diff
@@ -36,19 +43,76 @@ const _querySPSP = function * (receiver) {
...
+    spsp: spsp
+  }
+}
+
+/**
+  * Query an SPSP endpoint and get SPSP details
+  *
+  * @param {String} receiver webfinger account identifier (eg. 'alice@example.com') or URL to SPSP endpoint.
+  *
+  * @return {Promise<Object>} SPSP SPSP response from server
```
emschwartz: The SPSP response object should be documented

test/transportSpec.js
sharafian: I think tests like that would belong in some separate `cryptoSpec`. One problem, though, is that we don't have an implementation of `ilp` other than this one to generate fixtures.

emschwartz: 👍  on the idea of a `cryptoSpec`. I think having some fixtures generated manually using the steps described in the spec would be better than not having any at all

src/lib/spsp.js
sharafian: https://github.com/interledgerjs/ilp/blob/e9ff26cefedc513f48513c96997958984a3e85e6/src/lib/spsp.js#L221

src/lib/ipr.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+  * Create a payment request for use in the IPR transport protocol.
+  *
+  * @param {Object} params Parameters for creating payment request
+  * @param {String} params.destinationAmount Amount that should arrive in the recipient's account. This value is a string representation of an integer, expressed in the lowest indivisible unit supported by the ledger.
+  * @param {String} params.destinationAccount Target account's ILP address
+  * @param {String} params.receiverSecret Secret for generating IPR packets
+  * @param {String} [params.id=uuid.v4()] Unique ID for the request (used to ensure conditions are unique per request)
+  * @param {String} [params.expiresAt=30 seconds from now] Expiry of request
+  * @param {Object} [params.data=null] Additional data to include in the request
```
emschwartz: Is this really expecting an Object, or is this supposed to be a buffer?

src/lib/ipr.js
```diff
@@ -0,0 +1,69 @@
...
+  * @param {Object} [params.headers=null] Additional headers for private details
```
emschwartz: What kind of Object is this? A map with the headers as keys? If so it should say that

docs/README.template.md
```diff
@@ -23,30 +23,51 @@ A low-level JS <a href="https://interledger.org">Interledger</a> sender/receiver
...
 [snyk-url]: https://snyk.io/test/npm/ilp
 
-This is a low-level interface to ILP, largely intended for building ILP into other [Application layer](https://github.com/interledger/rfcs/tree/master/0001-interledger-architecture) protocols.
-
-#### The ILP Client includes:
+#### The ILP module includes:
 
 * [Interledger Payment Request (IPR)](#interledger-payment-request-ipr-transport-protocol) Transport Protocol, an interactive protocol in which the receiver specifies the payment details, including the condition
 * [Pre-Shared Key (PSK)](#pre-shared-key-psk-transport-protocol) Transport Protocol, a non-interactive protocol in which the sender creates the payment details and uses a shared secret to generate the conditions
+* [Simple Payment Setup Protocol (SPSP)](#simple-payment-setup-protocol-spsp), a higher level interface for sending ILP payments, which requires the receiver to have an SPSP server.
```
emschwartz: Please reorder these to be SPSP, PSK, then IPR

docs/README.template.md
```diff
@@ -55,50 +76,60 @@ This library handles the generation of payment requests, but **not the communica
...
 
+  sender.on('outgoing_fulfill', (transfer, fulfillment) => {
+    console.log(transfer.id, 'was fulfilled with', fulfillment)
+    stopListening()
+  })
+})()
 ```
 
-### Pre-Shared Key (PSK) Transport Protocol
+## Pre-Shared Key (PSK) Transport Protocol
```
emschwartz: We should just include a short description of PSK here and link to the spec

docs/README.template.md
```diff
@@ -125,55 +156,63 @@ from getting unwanted funds.
 ```js
 'use strict'
 
-const co = require('co')
-const ILP = require('.')
-const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
+import 'uuid'
```
emschwartz: The `uuid` module exports `v1` and `v4`

src/lib/spsp.js
```diff
@@ -150,56 +246,32 @@ const sendPayment = (plugin, payment, ilp) => {
   * @property {string} destination_account Receiver's ILP address.
   * @property {string} connector_account The connector's account on the sender's ledger. The initial transfer on the sender's ledger is made to this account.
   * @property {string} spsp SPSP response object, containing details to contruct transfers.
-  * @property {string} data extra data to attach to transfer.
+  * @property {Object} [publicHeaders={}] public headers for PSK data. The key-value pairs represent header names and values.
+  * @property {Object} [headers={}] headers for PSK data. The key-value pairs represent header names and values.
+  * @property {string} [data={}] extra data to attach to transfer.
```
emschwartz: SPSP data should be an arbitrary JSON object

src/utils/details.js
```diff
@@ -0,0 +1,196 @@
...
+  headers,
+  data
+}) {
+  const statusLineText = statusLine ? 'PSK/1.0\n' : ''
+  const headerLines = Object.keys(headers)
+    .map((k) => k + ': ' + headers[k])
+    .join('\n') + DATA_DELIMITER
+
+  let rawData = data
+  if (!Buffer.isBuffer(data) && typeof data === 'object') {
```
emschwartz: I might have originally argued for this behavior but I think it would be better for the transport layer to expect a buffer and reject/throw if you pass anything else. The SPSP layer should convert the JSON to a buffer

src/lib/psk.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+  * Create a payment request using a Pre-Shared Key (PSK).
+  *
+  * @param {Object} params Parameters for creating payment request
+  * @param {String} params.destinationAmount Amount that should arrive in the recipient's account. This value is a string representation of an integer, expressed in the lowest indivisible unit supported by the ledger.
+  * @param {String} params.destinationAccount Target account's ILP address
+  * @param {String} params.sharedSecret Shared secret for PSK protocol
+  * @param {String} [params.id=uuid.v4()] Unique ID for the request (used to ensure conditions are unique per request)
+  * @param {String} [params.expiresAt=30 seconds from now] Expiry of request
```
emschwartz: What do you think about making the default for PSK later? In some sense there's no reason to have an expiry at all..

sharafian: sure, I won't provide a default expiry for PSK.

docs/README.template.md
sharafian: v4 is the default, but I don't think it should matter either way, right?

emschwartz: Ah you're right, missed that

docs/README.template.md
```diff
@@ -55,50 +76,60 @@ This library handles the generation of payment requests, but **not the communica
...
 
+  sender.on('outgoing_fulfill', (transfer, fulfillment) => {
+    console.log(transfer.id, 'was fulfilled with', fulfillment)
+    stopListening()
+  })
+})()
 ```
 
-### Pre-Shared Key (PSK) Transport Protocol
+## [Pre-Shared Key (PSK) Transport Protocol](https://github.com/interledger/rfcs/blob/master/0016-pre-shared-key/0016-pre-shared-key.md)
```
emschwartz: Nitpick but PSK should come before IPR, like in the module outline

src/lib/spsp.js
```diff
@@ -150,56 +246,32 @@ const sendPayment = (plugin, payment, ilp) => {
   * @property {string} destination_account Receiver's ILP address.
   * @property {string} connector_account The connector's account on the sender's ledger. The initial transfer on the sender's ledger is made to this account.
   * @property {string} spsp SPSP response object, containing details to contruct transfers.
-  * @property {string} data extra data to attach to transfer.
+  * @property {Object} [publicHeaders={}] public headers for PSK data. The key-value pairs represent header names and values.
+  * @property {Object} [headers={}] headers for PSK data. The key-value pairs represent header names and values.
+  * @property {string} [data={}] arbitrary JSON object for additional data.
```
emschwartz: The description says it's a JSON object but the type says it's a string

src/lib/spsp.js
```diff
@@ -31,24 +38,100 @@ const _querySPSP = function * (receiver) {
...
-  const client = new IlpCore.Client(plugin, ilp)
-  const quote = yield client.quote({
+const quote = function * (plugin, {
+  receiver,
+  sourceAmount,
+  destinationAmount,
+  connectors,
+  id,
+  timeout
+}) {
```
emschwartz: If you pass in a memo here it won't actually be used so you'd need to first quote the payment params and then add the memo to the response. That doesn't seem great. (There should also be a test to make sure this works properly)

sharafian: I think that it's better to have the user set the `memo` field on the returned payment object, rather than potentially confusing them by having `quote` take an unrelated memo.

emschwartz:
LGTM once tests are passing and merge conflicts are resolved

PR #72, opened 2017-02-17
feat: BREAKING Use RejectionMessage for ILP Errors

sentientwaffle:
Depends on https://github.com/interledgerjs/ilp-plugin-bells/pull/98

PR #80, opened 2017-03-09
BREAKING RejectionMessage, new ilp-packet format

sentientwaffle:
PR includes #72

Depends on https://github.com/interledgerjs/ilp-core/pull/83


sentientwaffle:
@emschwartz I agree, but #79 should rebase onto this PR so that they won't conflict w/ each other at merge time.

PR #83, opened 2017-03-23
Fix shared secret size in SPSP

sharafian:
Adds tests for this behavior. Also includes a small fix to the PSK statusline regex that accepts minor version changes.


emschwartz:
Is there any reason not to allow a longer shared secret?

sharafian:
Force-merging because this is a blocker for @vhpoet, and because integration tests seem to be failing only because of a five-bells-integration-test bug.

PR #87, opened 2017-03-29
docs: fixes IPR example

dappelt:


dappelt:
@sharafian Would it be clear to the reader what we mean with _ledger's base unit_?

sharafian:
@dappelt That's the terminology that's used in the RFCs, but I don't think there's anywhere that defines it explicitly. There was talk of making a glossary RFC though, [here](https://github.com/interledger/rfcs/issues/180). For now I think it's enough if we just use language that's consistent within our own repos.

dappelt:
@sharafian Ok, I added a note.

PR #88, opened 2017-03-31
feat: Adapt to https://github.com/interledger/rfcs/pull/177

michielbdejong:


README.md
```diff
@@ -219,6 +219,30 @@ const receiver = new FiveBellsLedgerPlugin({
...
+$ npm run test
+```
+
+To run the integration tests, which test the integration between this node modules and related ones like
+[ilp-plugin-bells](https://github.com/interledgerjs/ilp-plugin-bells), run:
+
+```sh
+$ npm run integration-loader
+$ node ./node-modules/.bin/integration setup
+$ node ./node_modules/.bin/integration test
```
emschwartz: All of this is actually done by running `npm run integration`

emschwartz:
This should be marked as a breaking change because it will not work with older plugins

README.md
```diff
@@ -231,17 +231,13 @@ To run the integration tests, which test the integration between this node modul
...
 $ node ./node_modules/.bin/integration test
 ```
 
-To repeat one of the integration tests, run for instance:
-
-```sh
-$ cd integration-test
-$ mocha ../node_modules/five-bells-integration-test/src/tests/index.js
-```
+After a fix, you can repeat just the last step, or just the two last steps. It will complete faster than
```
emschwartz: I think you need to run `rm -rf integration-test/data` if you want to re-run `integration test`

README.md
michielbdejong: yeah, removed this again. we can now use the Docker image. :)

PR #89, opened 2017-03-31
Feat: add IPR encoding and decoding

sharafian:
Adds functions for the decoding and encoding of IPRs, as [this spec](https://github.com/interledger/rfcs/blob/master/asn1/InterledgerPaymentRequest.asn) defines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,8 @@ const { sharedSecret, destinationAccount } = ILP.PSK.generateParams({
   const { packet, condition } = ILP.PSK.createPacketAndCondition({
     sharedSecret,
     destinationAccount,
-    destinationAmount: '10', // denominated in the ledger's base unit
+    // denominated in the ledger's base unit
+    destinationAmount: '10',
   })
 
   const quote = await ILP.ILQP.quoteByPacket(sender, packet)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -394,10 +399,57 @@ Listen on a plugin for incoming PSK payments, and auto-generate fulfillments.
...
+| params.packet | <code>String</code> | ILP packet of this IPR |
+| params.condition | <code>String</code> | condition of this IPR |
+
+<a name="module_IPR..decodeIPR"></a>
+
+### IPR~decodeIPR(ipr) ⇒ <code>Object</code>
+Decode an IPR buffer for use in the IPR transport protocol
+
+**Kind**: inner method of <code>[IPR](#module_IPR)</code>  
+**Returns**: <code>Object</code> - Decoded IPR parameters, containing 'packet' and 'condition' as base64url strings.
```
emschwartz: Can you document the fields returned here?

src/lib/ipr.js
```diff
@@ -42,6 +46,79 @@ function createPacketAndCondition (params) {
...
+  assert(condition, 'params.condition must be specified')
+
+  const packetBuf = Buffer.from(packet, 'base64')
+  const conditionBuf = Buffer.from(condition, 'base64')
+
+  assert(conditionBuf.length === 32, 'params.condition must encode 32 bytes')
+
+  const writer = new oer.Writer()
+
+  writer.writeUInt8(IPR_VERSION)
```
emschwartz: @justmoon the ILP packets have a type and then the whole thing is in an OER octet string (so it has the length prefix before it) -- should IPR have either of those things?

src/lib/ipr.js
```diff
@@ -42,6 +46,79 @@ function createPacketAndCondition (params) {
...
+  * @param {Buffer} ipr encoded IPR buffer
+  *
+  * @return {Object} Decoded IPR parameters, containing 'packet' and 'condition' as base64url strings.
+  */
+function decodeIPR (ipr) {
+  assert(Buffer.isBuffer(ipr), 'ipr must be a buffer, got: ' + ipr)
+
+  const reader = oer.Reader.from(ipr)
+  const version = reader.readUInt8()
+  assert(version === 2, 'IPR version must be 2, got: ' + version)
```
emschwartz: This should be a constant declared at the top of the file

README.md
sharafian: What did you have in mind? Those are the only two fields returned.

emschwartz: True. It might be useful that has a method that does the opposite of `createIpr` and returns the same fields that one takes as parameters

PR #92, opened 2017-04-07
Feat: retry fulfillCondition automatically

sharafian:
If `fulfillCondition` fails in the `fulfill` callback, it will keep retrying until the transfer's expiry. If it still has not succeeded at that point, it will reject.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,9 +302,7 @@ SPSP query response
 | minimum_destination_amount | <code>string</code> | Integer string representing the minimum that the receiver will be willing to accept. |
 | ledger_info | <code>Object</code> | An object containing the receiver's ledger metadata. |
 | ledger_info.currency_code | <code>string</code> | The currency code of the receiver's ledger. |
-| ledger_info.currency_symbol | <code>string</code> | The currency symbol of the receiver's ledger. |
-| ledger_info.precision | <code>string</code> | The precision of the receiver's ledger. |
-| ledger_info.scale | <code>string</code> | The scale of the receiver's ledger. |
+| ledger_info.currency_scale | <code>string</code> | The currency scale of the receiver's ledger. |
 | receiver_info | <code>Object</code> | Additional information containing arbitrary fields. |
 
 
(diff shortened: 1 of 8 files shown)


src/lib/transport.js
```diff
@@ -137,7 +139,12 @@ function * listen (plugin, {
         destinationAccount,
         destinationAmount,
         fulfill: function () {
-          return plugin.fulfillCondition(transfer.id, fulfillment)
+          return retryPromise({
+            callback: () => plugin.fulfillCondition(transfer.id, fulfillment),
+            minWait: minFulfillRetryWait || 250,
```
emschwartz: Defaults should be declared as constants at the top of the file

src/lib/psk.js
```diff
@@ -21,7 +21,9 @@ const assert = require('assert')
   * @param {Buffer} [params.data=null] Additional data to include in the request
   * @param {Object} [params.headers=null] Additional headers for private PSK details. The key-value pairs represent header names and values.
   * @param {Object} [params.publicHeaders=null] Additional headers for public PSK details. The key-value pairs represent header names and values.
-  * @param {Object} [params.disableEncryption=false] Turns off encryption of private memos and data
+  * @param {Boolean} [params.disableEncryption=false] Turns off encryption of private memos and data
+  * @param {Number} [params.minFulfillRetryWait] Minimum amount of time to wait before retrying fulfillment
```
emschwartz: This should say what the default is

README.md
```diff
@@ -302,9 +302,7 @@ SPSP query response
 | minimum_destination_amount | <code>string</code> | Integer string representing the minimum that the receiver will be willing to accept. |
 | ledger_info | <code>Object</code> | An object containing the receiver's ledger metadata. |
 | ledger_info.currency_code | <code>string</code> | The currency code of the receiver's ledger. |
-| ledger_info.currency_symbol | <code>string</code> | The currency symbol of the receiver's ledger. |
-| ledger_info.precision | <code>string</code> | The precision of the receiver's ledger. |
-| ledger_info.scale | <code>string</code> | The scale of the receiver's ledger. |
```
emschwartz: Why is this commit also modifying the precision and scale?

test/utilSpec.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const chai = require('chai')
+const expect = chai.expect
```
emschwartz: Better to use either expect or assert but not both at the same time

README.md
sharafian: The commit that modified it just didn't build docs

src/lib/transport.js
```diff
@@ -137,7 +142,12 @@ function * listen (plugin, {
         destinationAccount,
         destinationAmount,
         fulfill: function () {
-          return plugin.fulfillCondition(transfer.id, fulfillment)
+          return retryPromise({
+            callback: () => plugin.fulfillCondition(transfer.id, fulfillment),
```
emschwartz: I still think we _might_ want to catch the final errors defined in the Ledger Plugin Interface so that we don't retry those

PR #97, opened 2017-04-24
Fix: clean up quote listeners after a quoting error

sharafian:
Currently, they're just being removed on a successful quote request, which leads to a memory leak. This makes it so `incoming_message` listeners are removed on an error response or on a timeout.


sharafian:
Force merging because license/cla is timing out.

PR #99, opened 2017-04-27
Feat: allow SPSP response to be passed in

sharafian:
Allows the SPSP response to be passed into `SPSP.quote`, potentially saving an HTTP request if the receiver has already been fetched. Also exports `validateSPSPResponse` from the ILP.SPSP module.

### DIFF
--- a/src/lib/spsp.js
+++ b/src/lib/spsp.js
@@ -46,6 +46,12 @@ const _querySPSP = function * (receiver) {
   return response
 }
 
+/**
+  * Validate a server's SPSP response, and throw an error if it's wrong.
+  *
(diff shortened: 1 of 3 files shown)


src/lib/spsp.js
```diff
@@ -108,17 +115,19 @@ const quote = function * (plugin, {
   destinationAmount,
   connectors,
   id,
-  timeout
+  timeout,
+  spspResponse
 }) {
   assert(plugin, 'missing plugin')
-  assert(receiver, 'receiver')
+  assert(receiver || spspResponse,
```
dappelt: Should there be an assertion that checks that `spspResponse` is a valid SPSP Response?

sharafian: fixed

PR #105, opened 2017-05-27
Feat: add listen all function for plugin factory

sharafian:
Allows you to listen on many plugins while just registering one event listener. Only compatible with the `ilp-plugin-bells` Factory class. `ILP.PSK.listenAll` takes the same receiver secret used for the `generateParams` function.

### DIFF
--- a/src/lib/ipr.js
+++ b/src/lib/ipr.js
@@ -146,10 +146,16 @@ function listen (plugin, params, callback) {
   return Transport.listen(plugin, params, callback, 'ipr')
 }
 
+// TODO: apidoc
+function listenAll (factory, params, callback) {
+  return Transport.listenAll(factory, params, callback, 'ipr')
+}
+
 module.exports = {
   createPacketAndCondition,
   createIPR,
   encodeIPR,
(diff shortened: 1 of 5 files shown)


src/lib/ipr.js
```diff
@@ -146,10 +146,16 @@ function listen (plugin, params, callback) {
   return Transport.listen(plugin, params, callback, 'ipr')
 }
 
+// TODO: apidoc
```
emschwartz: Can you add that?

test/transportSpec.js
```diff
@@ -598,5 +599,45 @@ data`, 'utf8')) }))
...
+      })
+
+      it('should listenAll', async function () {
+        this.callback = async function ({ fulfill }) {
+          await fulfill()
+        }
+
+        await Transport.listenAll(this.factory, this.params, this.callback)
+        
+        this.factory.emit('incoming_prepare', 'alice', this.transfer)
```
emschwartz: Can you show it fulfilling transfers for two different accounts?

PR #106, opened 2017-06-06
docs: fix example code in readme

michielbdejong:
Not sure which node version supports `import`? I tried with both v7 and v8, thought `require` is probably more common.

Inline calls to async functions apparently get parsed wrong if the semicolon of the previous statement is not there - I added it at the beginning of the line because then it's close to the syntax that needs it to be there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,15 +44,18 @@ provides a high-level interface:
 ```js
 'use strict'
 
-import { SPSP } from 'ilp'
-import FiveBellsLedgerPlugin from 'ilp-plugin-bells'
+const SPSP = require('ilp').SPSP
+const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
 
 const plugin = new FiveBellsLedgerPlugin({
   account: 'https://red.ilpdemo.org/ledger/accounts/alice',
   password: 'alice'
 })
 
-(async function () {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,15 +44,17 @@ provides a high-level interface:
...
+const FiveBellsLedgerPlugin = require('ilp-plugin-bells')
 
 const plugin = new FiveBellsLedgerPlugin({
   account: 'https://red.ilpdemo.org/ledger/accounts/alice',
   password: 'alice'
 })
 
-(async function () {
+;(async function () {
+  await plugin.connect()
```
michielbdejong: This was missing

README.md
```diff
@@ -112,15 +114,15 @@ const receiver = new FiveBellsLedgerPlugin({
 
 const receiverSecret = Buffer.from('secret_seed')
 const { sharedSecret, destinationAccount } = ILP.PSK.generateParams({
-  destinationAccount: receiver,
+  destinationAccount: 'us.usd.blue.bob',
```
michielbdejong: This was wrong

emschwartz: I think it's better to put `reciever.getAccount()` because otherwise it's unclear how you'd figure that address out

PR #109, opened 2017-07-19
fix: use 32 byte secrets (DO NOT MERGE YET)

adrianhopebailie:
- Change PSK to be 32 bytes (output of HMAC)
- Update tests to use known test vectors

@sharafian I have fixed the code and added test vectors to verify it.

There are some more complex test cases failing in the transportSpec which I don't understand well enough to debug.

Can I hand this over to you to add deprecation of 16 byte keys and try and get the test cases working again?


sharafian:
can this be closed?

adrianhopebailie:
It's so out of date I assume so

PR #112, opened 2017-08-03
fix: prevent line feed injection into headers

dappelt:


test/transportSpec.js
```diff
@@ -140,6 +140,13 @@ describe('Transport', function () {
       assert.equal(details.headers['header'], 'value')
     })
 
+    it('should throw an error if header value contains a line feed', function () {
+      this.params.headers = { 'Some-Header': 'value\nX-Header-Injection: Some-Value' }
```
michielbdejong: Add a test `this.params.headers = { 'Some-Header\nX-Header-Injection: Some-Value': 'value' }`

src/utils/details.js
```diff
@@ -15,6 +15,7 @@ function _createRequest ({
   headers,
   data
 }) {
+  validateHeaders(headers)
   const statusLineText = statusLine ? 'PSK/1.0\n' : ''
   const headerLines = Object.keys(headers)
     .map((k) => k + ': ' + headers[k])
```
michielbdejong: Maybe move this into the validateHeaders function, so it only loops over the object once.

src/utils/details.js
```diff
@@ -27,6 +28,15 @@ function _createRequest ({
   ])
 }
 
+function validateHeaders (headers) {
+  for (const name in headers) {
+    const value = headers[name].toString()
+    if (value && value.indexOf('\n') !== -1) {
```
michielbdejong: I would prefer to enforce a much stricter RegEx on both `name` and `value`, to allow only alpha-numerical values, that way you also disallow messy stuff like ':' in the name, `\r`, `\0`, `\del`, etc.

emschwartz: I think we need to allow unicode characters so that you can do things like send the sender's name (something ILP Kit is currently doing) in a header with non-ASCII characters.

test/transportSpec.js
```diff
@@ -140,6 +140,13 @@ describe('Transport', function () {
       assert.equal(details.headers['header'], 'value')
     })
 
+    it('should throw an error if header value contains a line feed', function () {
```
emschwartz: I would also add tests to make sure that you can include both URLs and Unicode characters (e.g. for names)

src/utils/details.js
dappelt: I think that would be too strict. This would for example not allow us to put URLs into a header.

src/utils/details.js
```diff
@@ -27,6 +30,15 @@ function _createRequest ({
   ])
 }
 
+function validateHeader (name, value) {
+  if (name.indexOf('\n') !== -1 || name.indexOf(':') !== -1) {
+    throw new Error(`Found forbidden characters in header name: ${name}`)
+  }
+  if (value && value.toString().indexOf('\n') !== -1) {
+    throw new Error(`Found forbidden characters in header value: ${value}`)
```
michielbdejong: I would do `${JSON.stringify(value)}`, to get:
```js
> throw new Error(`Found forbidden characters in header value: ${JSON.stringify(value)}`)
Error: Found forbidden characters in header value: "\rWarning: Your server is belongz to us!!!!! Pay the ransom using ILP,\nor we will wipe out your server logs:"
    at repl:1:7
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:336:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:533:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:191:7)
    at REPLServer.Interface._onLine (readline.js:238:10)
    at REPLServer.Interface._line (readline.js:582:8)
```
and not:
```js
> throw new Error(`Found forbidden characters in header value: ${value}`)
Warning: Your server is belongz to us!!!!! Pay the ransom using ILP,
or we will wipe out your server logs:
    at repl:1:128
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:336:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:533:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:191:7)
    at REPLServer.Interface._onLine (readline.js:238:10)
    at REPLServer.Interface._line (readline.js:582:8)
>

PR #121, opened 2017-12-03
chore: update dependencies and switch to async/await

justmoon:
* Switch to async/await
* Turn on linting for whole codebase, not just `src`
* Update eslint dependencies
* Fix lint issues
* Update all dependencies
* Add customizable connect timeouts (mocha 4 no longer force exits, so the tests were hanging waiting for a 10 second connect timeout)


src/lib/ilqp.js
```diff
@@ -186,17 +185,17 @@ function * quote (plugin, {
   })
 }
 
-function * quoteByPacket (plugin, packet) {
+async function quoteByPacket (plugin, packet, params) {
```
sublimator: Welcome to c# 2012 haha

PR #122, opened 2017-12-18
Update to LPI2

justmoon:
Builds on #121. Depends on interledgerjs/ilp-compat-plugin#7


src/lib/ilqp.js
```diff
@@ -56,26 +25,21 @@ function _serializeQuoteRequest (requestParams) {
...
   timeout
 }) {
-  const prefix = plugin.getInfo().prefix
+  plugin = compat(plugin)
   const requestPacket = _serializeQuoteRequest(quoteQuery)
   const requestType = requestPacket[0]
 
-  debug('remote quote connector=' + connector, 'query=' + JSON.stringify(quoteQuery))
+  debug('remote quote', 'query=' + JSON.stringify(quoteQuery))
   return plugin.sendRequest({
```
michielbdejong: Same question as https://github.com/interledgerjs/ilp-connector/pull/401#discussion_r157580607

michielbdejong: So then is this PR intended to preserve or break support for ILQP?

test/mocks/mockPlugin.js
```diff
@@ -22,15 +23,28 @@ module.exports = class MockPlugin extends EventEmitter2 {
...
+    }
+    this._handler = handler
+  }
+
+  deregisterTransferHandler (handler) {
+    this._handler = null
   }
 }
+
+MockPlugin.version = 2
```
sentientwaffle: Should this be `lpiVersion` (see https://github.com/interledgerjs/ilp-compat-plugin/blob/d8331126a92ad2b65a7e7c4513ce44c13d485340/index.js#L31)?

test/mocks/mockPlugin.js
```diff
@@ -22,15 +23,28 @@ module.exports = class MockPlugin extends EventEmitter2 {
     return 'test.example.alice'
   }
 
-  sendRequest () {
+  sendRequest (request) {
```
sentientwaffle: If this is mocking a LPI2 plugin, `sendRequest` should be removed.

src/lib/ilqp.js
emschwartz: Preserve support for the moment, even though it's a deprecated feature

michielbdejong: I think it would be more precise to say that we are experimenting with the idea of deprecating ILQP, not that we decided to deprecate it?

emschwartz: Maybe, but if someone were going to write a new implementation of ILP right now I wouldn't recommend implementing ILQP

src/lib/ildcp.js
```diff
@@ -0,0 +1,45 @@
...
+const debug = require('debug')('ilp:ildcp')
+
+const PEER_PROTOCOL_CONDITION = Buffer.from('Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=', 'base64')
+const PEER_PROTOCOL_EXPIRY_DURATION = 60000
+
+const getAccount = async (plugin) => {
+  plugin = compat(plugin)
+
+  const data = await plugin.sendData(IlpPacket.serializeIlpPrepare({
+    // TODO: Should be zero, but our plugins suck
```
emschwartz: lol

...we should probably make them not suck because the connector will actually fulfill this

src/lib/ildcp.js
```diff
@@ -0,0 +1,45 @@
...
+    amount: '1',
+    executionCondition: PEER_PROTOCOL_CONDITION,
+    expiresAt: new Date(Date.now() + PEER_PROTOCOL_EXPIRY_DURATION),
+    destination: 'peer.config',
+    data: Buffer.alloc(0)
+  }))
+
+  console.log('data', data)
```
emschwartz: Remove console.log statement

src/lib/ilqp.js
```diff
@@ -55,41 +32,32 @@ function _serializeQuoteRequest (requestParams) {
...
     }
-    if (isErrorPacket || responseType === requestType + 1) {
-      return Object.assign({responseType}, packetData)
+    const packetData = IlpPacket.deserializeIlpPacket(ilp).data
+
+    if (ilp[0] === IlpPacket.Type.TYPE_ILP_REJECT) {
+      debug('remote quote error. ilpError=%j', packetData)
     }
-    throw new Error('Quote response packet has incorrect type')
+    console.log('packetData', packetData)
```
emschwartz: Remove console.log statements

src/errors/receiver-rejection-error.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+const BaseError = require('extensible-error')
+const { codes } = require('../utils/ilp-errors')
+
+class ReceiverRejectionError extends BaseError {
+  constructor (message) {
+    super(message)
+
+    this.ilpErrorCode = codes.F00_BAD_REQUEST
```
emschwartz: Shouldn't this be an `F99: Application Error`?

src/lib/transport.js
```diff
@@ -155,134 +105,146 @@ async function listenAll (factory, {
...
 }) {
-  const account = plugin.getAccount()
   const receiverId = base64url(cryptoHelper.getReceiverId(receiverSecret))
 
   if (!transfer.executionCondition) {
     debug('notified of transfer without executionCondition ', transfer)
-    return _reject(plugin, transfer.id, ilpErrors.F00_Bad_Request({
-      message: 'got notification of transfer without executionCondition'
-    }))
+    throw new BadRequestError('got notification of transfer without executionCondition')
```
emschwartz: This isn't possible anymore

src/lib/transport.js
```diff
@@ -155,134 +105,146 @@ async function listenAll (factory, {
...
-  // TODO: should this just be included in this function?
-  const err = await _validateOrRejectTransfer({
-    plugin,
-    transfer,
-    allowOverPayment,
-    receiverSecret
-  })
+  try {
+    const transfer = IlpPacket.deserializeIlpPrepare(packet)
+    const parsed = IlpPacket.deserializeIlpPayment(transfer.data)
```
emschwartz: Putting the old packet type inside the new packet type is a pretty strange hack to make this change less breaking, so it would definitely be good to include a comment explaining this

src/lib/ildcp.js
justmoon: Hehe, this comment was from a previous version. No longer applicable because in ilp4 the plugin doesn't even know the amount. (sendData just sends opaque data and for amount=0 you just don't sendMoney afterwards.)

justmoon:
@emschwartz Addressed your comments, still need to fix tests.

emschwartz:
Looks good, can you also update the README explanation of how to use it?

src/lib/ipr.js
```diff
@@ -136,29 +137,34 @@ function createIPR (params) {
...
+  * @param {Buffer} params.receiverSecret Secret to generate fulfillments with
   * @param {Buffer} [params.allowOverPayment=true] Accept payments with higher amounts than expected
   * @param {IncomingCallback} callback Called after an incoming payment is validated.
   *
   * @return {Object} Payment request
   */
-function listen (plugin, params, callback) {
-  return Transport.listen(plugin, params, callback, 'ipr')
+async function listen (plugin, params, callback) {
+  const address = (await ILDCP.fetch(plugin.sendData.bind(plugin))).clientAddress
```
sentientwaffle: `plugin.sendData.bind` will fail for v1 plugins since `plugin` hasn't been wrapped by `ilp-compat-plugin`.

PR #126, opened 2018-01-13
Feat: add payment-pointer SPSP identifiers

sharafian:
Following discussions with @justmoon and @emschwartz , we need a way to identify people on Interledger. The current SPSP identifiers are indistinguishable from email addresses, and also require a cumbersome webfinger endpoint.

This is a proposal to add SPSP identifiers in the form of `$foo.com` to commonly identify people on Interledger. The current form of `bob@example.com` would become `$bob.example.com` or `$example.com/bob`. HTTPS is always used, so these would become `https://bob.example.com/` and `https://example.com/bob` respectively. The request is sent with `Accept: application/x-spsp-response`, so that one could run a website on `example.com` and still use `$example.com` to serve SPSP responses.

My proposed name for this identifier is a "Payment Pointer," but I'm open to other suggestions.

The reasoning to do this over `user@example.com` is:

1. These identifiers are easily recognizable as payment pointers. They won't be easily confused with email addresses or HTTP endpoints.
2. The `/^\$[A-Za-z0-9\-]+\.[A-Za-z0-9\-.]+$/` regex (or a more restrictive one) is specific enough that browsers or extensions could easily implement behavior for these payment pointers.
3. It's human readable and more visually pleasing than an `spsp://` URI or something similar. Also easier to parse than a cryptocurrency address.
4. Continues to leverage existing web infrastructure.

cc @emschwartz @justmoon @michielbdejong @adrianhopebailie


michielbdejong:
One question is whether Alice continues to type 'bob@blue.ilpdemo.org' into ilp-kit's GUI or (this would be a UI change) from now on should type '$bob.blue.ilpdemo.org' or '$blue.ilpdemo.org/bob' into that same input field. I could answer that question but that would make me complicit in bike-shedding. ;)

Another point I wanted to raise is that we should distinguish users from invoices. An invoice is a virtual trustline between Bob and Alice, which has a balance. It is not uniquely defined in the context of Bob, but in the combination of Bob with Alice. So if Charlie asks Bob for a payment point, would he get his own Bob-and-Charlie invoice, with a balance? Or would he get the same "this is Bob" information as Alice?

sharafian:
> One question is whether Alice continues to type 'bob@blue.ilpdemo.org' into ilp-kit's GUI or (this would be a UI change) from now on should type '$bob.blue.ilpdemo.org' or '$blue.ilpdemo.org/bob' into that same input field. I don't have a strong preference one way or another.

I think whenever we upgrade ILP Kit to use ILPv4 we can also upgrade it to use SPSPv3 and payment pointers

> An invoice is a virtual trustline between Bob and Alice, which has a balance. It is not uniquely defined in the context of Bob, but in the combination of Bob with Alice.

I think an invoice is more like a bucket to be filled, where anyone can pay into it. So it would be identified only by payment-pointer, which in turn is resolved to a specific SPSP receiver that tracks the amount incoming. Alice can give her invoice payment pointer to Bob, and Bob could either pay the whole thing, give it to Charlie to pay on his behalf, or any number of people could pay in any proportion. The only thing Alice cares about is that the invoice is filled.

michielbdejong:
Are you just proposing we change how things work under the hood, or also proposing a change in the user-facing functionality? Because ilp-kit does not currently deal with invoices, right?

adrianhopebailie:
I like where this is going but I think it over-simplifies the issue. These identifiers (pointers) are a UX thing and the choice of one over the another infers information not to mention that there are few layers of indirection here (which we can maybe clear away?)

E.g. If I give you my bank account details as a way to pay me then you can infer that you need to pay me into a bank account. If I give you an email address I need to tell you that you can send a PayPal payment to that address as opposed to using it for current SPSP payments.

I assume that If I give you the identifier `$bob.blue.ilpdemo.org` you should you infer that I can be paid via ILP and that you should use SPSP service to discover how?

Critically, how do I infer what the host portion is vs the account portion? i.e. Why is `$bob.blue.ilpdemo.org` not the same as `bob.blue@ilpdemo.org`

Have you seen Square's CashTags: https://squareup.com/help/us/en/article/5504-how-to-claim-a-cashtag

An interesting thought is to take that idea and make it a standard. What if all pointers take the form `account$host`? 

That way CashTags are still usable if expressed as part of the `cash.me` domain. i.e. Existing Square CashTags are expressed as `user$cash.me` when exchanging them as a payment pointer.

So the discovery flow is simple:
Bob -> Alice: "Here's my payment pointer `bob$blue.ilpdemo.org`"
Alice: SPSP request to `https://blue.ilpdemo.org/$bob` with appropriate headers to indicate this is a request for SPSP JSON data (`Accept: application/spsp+json`)

sharafian:
> Critically, how do I infer what the host portion is vs the account portion? i.e. Why is $bob.blue.ilpdemo.org not the same as bob.blue@ilpdemo.org

I'm not sure I understand what you mean. The mapping is just `$example.com/path` -> `https://spsp.example.com/path`. Which part of that is the "account" is up to whatever server is listening on that endpoint

adrianhopebailie:
> The mapping is just $example.com/path -> https://spsp.example.com/path

I misunderstood. I thought you wanted the mapping to be `$bob.example.com` -> `https://example.com/bob`.

As discussed in https://github.com/interledger/rfcs/pull/367 I think this is a bad idea. There are many reasons for not having hardcoded subdomain prefixes (one of which you have discovered already in the complexities around SSL certs).

sharafian:
I believe this is deprecated in favor of https://github.com/interledgerjs/ilp-protocol-psk2 and https://github.com/sharafian/ilp-protocol-spsp

PR #134, opened 2018-03-28
[WiP] Add support for Loopback Transport

michielbdejong:
work in progress...


michielbdejong:
Example usage:
```sh
$ node test.js
{ destinationAmount: '10' }
{ result: true }
```

michielbdejong:
@sharafian still some unfinished things on the todo list, but would appreciate some early feedback on this PR

scripts/rouletteSender.js
```diff
@@ -0,0 +1,41 @@
+const PluginBtp = require('ilp-plugin-btp')
+const ILP = require('..')
+
+;(async function test () {
+  const plugin1 = new PluginBtp({ server: 'btp+ws://:plugin1@localhost:8912/' })
+  const plugin2 = new PluginBtp({ server: 'btp+ws://:plugin2@localhost:8912/' })
+  await plugin1.connect()
+  await plugin2.connect()
+  const loop = await ILP.LT.createLoop(plugin1, plugin2)
```
sharafian: Does this send plugin1 to plugin2? It might be clearer if we made this an object with names for the two plugins (i.e. outgoing vs. incoming, or sender vs. receiver), rather than using positional arguments.

scripts/rouletteSender.js
```diff
@@ -0,0 +1,41 @@
...
+    let destinationAmountSeen
+    const result = await loop.pay({
+      sourceAmount: '1',
+      expiresAt: new Date(new Date().getTime() + 10000),
+      loopbackHandler: (destinationAmount) => {
+        destinationAmountSeen = parseInt(destinationAmount)
+        cummSeen += destinationAmountSeen
+        numSeen++
+        // console.log({ destinationAmountSeen, threshold: (cummSeen / numSeen) })
+        return (destinationAmountSeen > (cummSeen/numSeen))
```
sharafian: I don't think you should use this equation in practice, because then if the rate goes down the payment won't be completed until the rate goes up again (which may never happen).

What code would one usually run here? And do we want to have some convenience method to send a chunked payment that does all of this automatically? I do like that you can specify the per-chunk logic manually too, though

src/lib/lt.js
```diff
@@ -0,0 +1,64 @@
...
+
+  async pay ({ sourceAmount, expiresAt, loopbackHandler }) {
+    const fulfillment = crypto.randomBytes(32)
+    const executionCondition = sha256(fulfillment)
+    const packet = IlpPacket.serializeIlpPrepare({
+      amount: sourceAmount,
+      expiresAt,
+      executionCondition,
+      destination: this.destination,
+      data: Buffer.from([])
```
sharafian: There may still be a reason to include data here. What if the sender wants to correlate packets between the sending and receiving sides by putting some kind of ID in the data?

test/lt.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+const IlpPacket = require('ilp-packet')
+const chai = require('chai')
+const assert = chai.assert
+const LT = require('..').LT
+const MockPlugin = require('./mocks/mockPlugin')
+const chaiAsPromised = require('chai-as-promised')
+chai.use(chaiAsPromised)
+
+describe('IT', function () {
```
sharafian: LT?

scripts/rouletteSender.js
michielbdejong: Yeah, and even if it stays constant, no chunks ever get accepted :) I'll improve the expression for the demo.

Rather than a convenience function, maybe we can give a few examples, or a parameterized expression where you specify a trade-off between exchange rate and speed under some assumptions about the distribution of exchange rates. If the sender has multiple ILSPs, it would also make sense to send chunks over multiple paths, to maximize exploration.

src/lib/lt.js
michielbdejong: Yeah, I had that at first, but then realized that the condition is already a unique identifier (and is also used actually used as such in the code, as the index to the `pending` hashmap).

test/lt.test.js
michielbdejong: Loopback Transport (see https://github.com/interledger/rfcs/pull/416/files#diff-ce1d0a4db7a75b9030d0436e521d092bR115)

sharafian: sorry, I meant that you wrote `IT` there and I think it should be `LT`

README.md
```diff
@@ -36,6 +38,34 @@ The Javascript client library for <a href="https://interledger.org">Interledger<
...
 
+## [Loopback Transport (LT)](https://github.com/interledger/rfcs/pull/416)
+
+```js
+const PluginBtp = require('ilp-plugin-btp')
+const ILP = require('.')
+
+;(async function test () {
+  const pluginOut = new PluginBtp({ server: 'btp+wss://:plugin1@amundsen.ilpdemo.org:1810/' })
+  const pluginIn = new PluginBtp({ server: 'btp+wss://:plugin2@amundsen.ilpdemo.org:1810/' })
```
sharafian: should this use a plugin if we're expecting people to also use that BTP channel for transmitting data? The plugin doesn't expose any way to send other than `sendData`, which puts things into the ILP protocol data. So the full capability of BTP isn't exposed to the user.

The spec for LT also says that it uses BTP or HTTP-OER-ILP, but this implementation allows any plugin to be used. Maybe instead of taking a plugin for pluginIn, we should pass in a URL. A plugin may be constructed internally but the BTP connection could at least be exposed to the user. Unfortunately we'd need an API for that BTP connection because we can't expect the user to serialize messages manually.

michielbdejong: > should this use a plugin if we're expecting people to also use that BTP channel for transmitting data? 

This is just example code, it also has the amount `'10'` hard-coded. You could use the LT implementation from this PR using an augmented version of PluginBTP, which has an extra `sendCustomData` method. I do think, at least for now, we should implement LT here using LPIv2 plugins. If we see later that we want to create an LPIv3, we can always update this code again.

> this implementation allows any plugin to be used

Correct, because neither LT nor ILP put any restrictions on the ledger layer. 

> Maybe instead of taking a plugin for pluginIn, we should pass in a URL

Yes, but that would then be the payment pointer (i.e., the LSD URL). So you could do something like:

```js
const pluginOut = await ILP.LSD.setupLoopbackPlugin(`$example.com/bob`)
const loop = ILP.LT.createLoop({ pluginIn, pluginOut })
```

But this PR only implements that second step.

michielbdejong:
This PR now relies on https://www.npmjs.com/package/ilp-protocol-lt which can also be used as a stand-alone library.

We can leave this PR dormant until we decide whether we want to update the ILP module as a whole, or deprecate it.

PR #135, opened 2018-08-21
BREAKING: Upgrade to ilpv4

adrianhopebailie:


package.json
```diff
@@ -1,18 +1,14 @@
 {
   "name": "ilp",
-  "version": "12.0.1",
+  "version": "14.0.0",
```
sentientwaffle: Why skip `13.0.0`?

kincaidoneil:
@adrianhopebailie What utility does this repo provide over just requiring those modules individually? I think it's confusing and we should consider archiving it.

package.json
adrianhopebailie: Unlucky 13 😄

adrianhopebailie:
> What utility does this repo provide over just requiring those modules individually? I think it's confusing and we should consider archiving it.

I considered that but I think having a single library to use from an application is valuable. The goal with the `ilp` package was always convenience for developers using ILP in applications (as opposed to writing connectors or plugins).

It's likely that we'll add some "application" layer functions here too in future that abstract away STREAM.

@emschwartz WDYT?

kincaidoneil:
I see what you mean. My thinking is, the bigger impediment to using ILP in applications is not the inconvenience of using multiple libraries, but rather that there's no guide or documentation instructing how they work together. (And I think the outdated README here could contribute to that confusion.) That said, it sounds like adding some abstractions on top of STREAM could provide some utility.

emschwartz:
I'm leaning towards getting rid of this module. Originally it made sense because we had multiple recommended transport protocols, but now there's only one. It doesn't really make sense to bundle application layer protocols because by definition they'll be used by pretty different use cases and you probably won't ever need two of them at the same time. I agree with @kincaidoneil that the bigger problem is the lack of or out of date documentation about what you should be using.

adrianhopebailie:
Agree that the bigger problem is documentation but I see this lib as a solution not another problem. Giving developers a single place to to "get the ILP code and docs" is, to my mind the solution tot hat problem.

What I want to avoid is making developers go to 5 different places to piece everything together. Docs on the website that tell you to get X lib for this and Y lib for that etc.

Happy to fix the README here and make it the start of a developer "Getting Started Guide" before merging and releasing a new version of the package. The advantage is that this README is what they see when they find the package on npm (which is where a lot of developers will go first).

adrianhopebailie:
> If we go the route of bundling libraries in this module we should include Cog as well

@sharafian my initial reaction was, yes! But on reviewing Cog I think there is a bit of cleanup required and some tests before I'd pull it in.

I'll log as an issue to track but that should definitely be the end-goal.

